PR #90, opened 2015-01-01
Enh/nidm results [final]

infocortex:
feature complete

still todo:
[√] migration test
[√] addditional unit test(s)


chrisfilo:
Migration went well, test pass. However edit images dialog does not work. Clicking on any of the images does not change anything. Neighter does clicking on the Add a new image button. JS console gives the following error:

	Uncaught TypeError: Cannot read property 'slice' of nullj query-2.1.0.min.js:2

chrisfilo:
Upload of .nidm.zip through "Upload Archive" works well. NIDM files are displayed well and editing individual files and maps works ok. However, pycortex does not work on the collection level. The loading indicator is spinning infinitely and dev console gives the following error:

	Failed to load resource: the server responded with a status of 404 (NOT FOUND) http://127.0.0.1:8000/media/images/3/spm_example.nidm/pycortex_all/fsaverage_[inflated]_mg2_9_['rois'].json

infocortex:
I will try to reproduce with a fresh migration.  Sounds like maybe JS isn't
loading, did you do .manage.py collectstatic after merging?

On Wed, Dec 31, 2014 at 5:01 PM, Chris Filo Gorgolewski <

> Migration went well, test pass. However edit images dialog does not work.
> Clicking on any of the images does not change anything. Neighter does
> clicking on the Add a new image button. JS console gives the following
> error:
>
> Uncaught TypeError: Cannot read property 'slice' of nullj query-2.1.0.min.js:2
>
> —
> <https://github.com/NeuroVault/NeuroVault/pull/90#issuecomment-68477237>.
>



-- 
*Gabriel Rivera*

InfoCortex UG
rivera@infocortex.com

+49 (0) 176 7223 1002
+1 917.375.0045

Friedensstraße 11/Kaiserplatz
D-60311 Frankfurt am Main
www.infocortex.com

chrisfilo:
I did collectstatic, but it did not help with neither of the bugs. The ne tests fails with:

	FAIL: testUploadNIDMZip (neurovault.apps.statmaps.tests.test_nidm.NIDMResultsTest)
	----------------------------------------------------------------------
	  File "/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tests/test_nidm.py", line 109, in testUploadNIDMZip
	    self.assertEquals(map_img.name,info['output_row']['name'])
	AssertionError: u'Shad FStatistic' != u'Generation FStatistic'
	- Shad FStatistic
	+ Generation FStatistic

infocortex:
fixed the failing test in bce5c28.  I made a completely fresh vagrant build and migrated from master to the branch, and the javascript is all working for me.  maybe you are getting a cached copy?  dev tools open?

I will test the pycortex issue next, afaik everything else is good to go, except for the aforementioned.

infocortex:
pycortex bug fixed in  c9f74ab.

chrisfilo:
The dialogs started working after getting a new vm instance. I tried this NIDM file: https://03b635f15995375276571521c484a221c331857d.googledrive.com/host/0B77zr9yIiKOTM19GRkt2d3E0TXc/two_contrasts.nidm.zip and it gives me 18 statistical maps instead of 6. Probably there is a bug in the SPARQL query. IT would be good to include this file in tests (and test for number of maps).

chrisfilo:
I can confirm that the pycortex bug is fixed.

chrisfilo:
I found one more thing. I cannot resolve URLs from the .ttl file. That applies to all files:

	Page not found (404)
	Request Method:	GET
	Request URL:	http://127.0.0.1:8000/collections/1/two_contrasts.nidm/FStatistic_F001.nii.gz
	"/opt/image_data/images/two_contrasts.nidm/FStatistic_F001.nii.gz" does not exist

infocortex:
Regarding the sparql query issue with `two_contrasts.nidm.zip`, the query returns not only redundant results for each contrasts, which isn't a problem, but also conflicting names for the same map type and file, see 'Shad' vs. 'Generation':

```
12:Generation nidm#TStatistic /tmp/tmp3ak8oN/nidm/TStatistic_T001.nii.gz
12:Generation nidm#FStatistic /tmp/tmp3ak8oN/nidm/FStatistic_F001.nii.gz
12:Generation nidm#TStatistic /tmp/tmp3ak8oN/nidm/TStatistic_T002.nii.gz
12:Generation nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_T001.nii.gz
12:Generation nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_T002.nii.gz
12:Generation nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_F001.nii.gz
12:Generation nidm#TStatistic /tmp/tmp3ak8oN/nidm/TStatistic_T001.nii.gz
12:Generation nidm#FStatistic /tmp/tmp3ak8oN/nidm/FStatistic_F001.nii.gz
12:Generation nidm#TStatistic /tmp/tmp3ak8oN/nidm/TStatistic_T002.nii.gz
12:Generation nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_T001.nii.gz
12:Generation nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_T002.nii.gz
12:Generation nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_F001.nii.gz
6:Shad nidm#TStatistic /tmp/tmp3ak8oN/nidm/TStatistic_T001.nii.gz
6:Shad nidm#FStatistic /tmp/tmp3ak8oN/nidm/FStatistic_F001.nii.gz
6:Shad nidm#TStatistic /tmp/tmp3ak8oN/nidm/TStatistic_T002.nii.gz
6:Shad nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_T001.nii.gz
6:Shad nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_T002.nii.gz
6:Shad nidm#ZStatistic /tmp/tmp3ak8oN/nidm/ZStatistic_F001.nii.gz
```

6 incidences of the name 'Shad', and 12 of 'Generation' ...the name conflict makes it impossible to unique-ify the list at the moment.  

I do have this file in the test suite, but its just whitelisted at the moment so the test passes.  I haven't looked into the query itself yet...also worth noting is the query runs very slow on this file.

chrisfilo:
It might be worth using "rdfs:label" instead of "nidm:contrastName" for naming. This will solve the issue for this .nidm.zip file. However there is nothing in the standard preventing having two statistical maps with identical labels. The only unique aspect (which you might want to use in the query) are the entity ids for example "niiri:0180617f-bfb4-4b83-9c08-319bd61988fc". We'll need yo use them in case labels clash. I hope this helps.

Regarding performance - I also experienced slowness (which is ridiculous considering the size of the .ttl file). Maybe @cmaumet, @dbkeator or @satra could help. This is the query we are talking about: https://github.com/infocortex/NeuroVault/blob/enh/nidm_results/neurovault/apps/statmaps/nidm_results.py#L65

infocortex:
Fixed serving file contents shown in .ttl:  b7c51df

chrisfilo:
Yup it works now. The duplicate entries are the only remaining issue.

infocortex:
OK, I switched to using the rdf:label as the basis for the names (adding the map type if the label is not unique), and ultimately filtering the query output by unique file names, which I think is sensible since the unique nii.gz's are what we're after: 1ce8455

Although something dubious is going on with that file, I think this sorts out our requirement.  One possible caveat is that the labels tend to have the map type in the name, which is slightly redundant in the UI, perhaps.

chrisfilo:
That looks reasonable. What do you mean when you say that "something dubious" is going on with that file?

infocortex:
I just meant that I suspected that file might be invalid or non-conformant, but I don't know enough about rdf to say for sure.  In any case, I think we have the needed handling in the parser.

One other quick note- @tyarkoni 's bootstrap3 upgrade #93  will require some manual merging against this branch (I rewrote some haml and coffeescript back to plain html/js, and used a small amount of boostrap v.2 specific css, so its probably easiest to merge that pr first and I can pull the needed changes into here.

cmaumet:
@chrisfilo, @infocortex: sorry for my late reply. Regarding the `two_contrasts.nidm.zip` example, I would not be surprised if this there were inaccuracies in the values of the `contrastName` attribute for this particular example. I would need to re-create an example for you with the latest version of the exporter.

In an FSL NIDM-Results export, you would expect to find the same `contrastName` for two statistic map: the `TStatistic` (or `FStatistic`) and the corresponding `ZStatistic`. While the standard does not prevent to have the same `contrastName` for different contrasts, I would say that this should be avoided. I hope this helps.

PR #97, opened 2015-01-06
[hotfix] edit images navigation stuff

### LINKED ISSUE #95: Problems with add/edit images dialog
Steps to replicate:
1. create a new collection
2. press edit images
3. press add statistic map (don't fill any fields)
4. press add atlas (don't fill any fields)
5. select first "[new image]" from the list 
6. select second "[new image]" from the list

results the form on the right does not change, selecting anything on the list does not seem to have any influence. Similar behaviour happens when editing a collection that already has some files

infocortex:
fixed navigation for new images in formset, also name updates in navigation. fixes #95

### DIFF
--- a/neurovault/apps/statmaps/static/scripts/clone-form.js
+++ b/neurovault/apps/statmaps/static/scripts/clone-form.js
@@ -1,4 +1,4 @@
-var active_form, showForm, updateImageName, cloneMore;
+var active_form, showForm, updateImageName, cloneMore, mapNavLink, getLastTextNode;
 
 active_form = 0;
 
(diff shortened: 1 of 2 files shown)


chrisfilo:
Now when I edit images on an empty collection I get a empty form that is neither a statistic map nor atlas or NIDM Results form. Maybe we should go back to having "None" that disappears as soon as user adds one image?

infocortex:
Ah, right.  How about:

- a notification box  ("This collection contains no images.  Use the 'Add
Images' button to add content to this collection").

- A [No images] in the navigation on the left.

And/ or:

- a blank statistic map image (instead of the blank un-typed image...)


On Tue, Jan 6, 2015 at 10:32 AM, Chris Filo Gorgolewski <

> Now when I edit images on an empty collection I get a empty form that is
> neither a statistic map nor atlas or NIDM Results form. Maybe we should go
> back to having "None" that disappears as soon as user adds one image?
>
> —
> <https://github.com/NeuroVault/NeuroVault/pull/97#issuecomment-68909369>.
>



-- 
*Gabriel Rivera*

InfoCortex UG
rivera@infocortex.com

+49 (0) 176 7223 1002
+1 917.375.0045

Friedensstraße 11/Kaiserplatz
D-60311 Frankfurt am Main
www.infocortex.com

chrisfilo:
would go with notification box on the right that would force users to make a choice on the left

infocortex:
I had to add an additonal fix, but this is done (for real this time...)

PR #98, opened 2015-01-06
Fix upload folder period

jbwexler:


neurovault/apps/statmaps/views.py
```diff
@@ -375,7 +375,7 @@ def upload_folder(request, collection_cid):
                             dom = minidom.parse(os.path.join(root, fname))
                             for atlas in dom.getElementsByTagName("summaryimagefile"):
                                 print "found atlas"
-                                path, base, ext = split_filename(atlas.lastChild.nodeValue)
+                                path, base = os.path.split(atlas.lastChild.nodeValue)
```
chrisfilo: does this really work? Won't it trip on .nii.gz files?

jbwexler: actually atlas.lastChild.nodeValue doesn't have an extension. ex: '/HarvardOxford/HarvardOxford-cort-maxprob-thr25-2mm'

chrisfilo: If that's the case this change should not make any difference...

jbwexler: the problem was for atlases whose base had a period in it. in those cases it would divide the part before and after the period and call them base and ext, but only base is used later so the stuff after the period was lost

PR #102, opened 2015-01-12
FEAT directory support

infocortex:
works well with original example.  importing the converter class seems preferable to calling the executable, and gives us control over the exception handling in our process.

todo:
 - testing
 - etc

### DIFF
--- a/neurovault/apps/statmaps/utils.py
+++ b/neurovault/apps/statmaps/utils.py
@@ -368,37 +368,3 @@ def get_server_url(request):
         return request.META['HTTP_ORIGIN']
     urlpref = 'https://' if request.is_secure() else 'http://'
     return '{0}{1}'.format(urlpref,request.META['HTTP_HOST'])
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
--- a/neurovault/apps/statmaps/utils.py
+++ b/neurovault/apps/statmaps/utils.py
@@ -289,22 +289,24 @@ def populate_nidm_results(request,collection):
     return form.instance
 
 
-def populate_feat_directory(request,collection):
+def populate_feat_directory(request,collection,existing_dir=None):
     from nidmfsl.fsl_exporter.fsl_exporter import FSLtoNIDMExporter
-    tmp_dir = tempfile.mkdtemp()
+    tmp_dir = tempfile.mkdtemp() if existing_dir is None else existing_dir
     exc = ValidationError
 
     try:
-        zip = zipfile.ZipFile(request.FILES['file'])
-        zip.extractall(path=tmp_dir)
+        if existing_dir is None:
+            zip = zipfile.ZipFile(request.FILES['file'])
+            zip.extractall(path=tmp_dir)
 
-        rootpaths = [v for v in os.listdir(tmp_dir) if not v.startswith('.')]
+        rootpaths = [v for v in os.listdir(tmp_dir)
+                     if not v.startswith('.') and not v.startswith('__MACOSX')]
         if not rootpaths:
-            raise exc("No contents found in the FEAT zip file.")
+            raise exc("No contents found in the FEAT directory.")
         subdir = os.path.join(tmp_dir,rootpaths[0])
-        feat_dir = subdir if len(rootpaths) and os.path.isdir(subdir) else tmp_dir
+        feat_dir = subdir if len(rootpaths) is 1 and os.path.isdir(subdir) else tmp_dir
     except:
-        raise exc("Unable to unzip the FEAT zip file: \n{0}.".format(get_traceback()))
+        raise exc("Unable to unzip the FEAT directory: \n{0}.".format(get_traceback()))
     try:
         fslnidm = FSLtoNIDMExporter(feat_dir=feat_dir, version="0.2.0")
         fslnidm.parse()
@@ -316,7 +318,12 @@ def populate_feat_directory(request,collection):
         raise exc("Unable find nidm export of FEAT directory.")
 
     try:
-        destname = request.FILES['file'].name.replace('feat.zip','feat.nidm.zip')
+        if existing_dir is not None:
+            dname = os.path.basename(feat_dir)
+            suffix = '.nidm.zip' if dname.endswith('feat') else '.feat.nidm.zip'
+            destname = dname + suffix
+        else:
+            destname = request.FILES['file'].name.replace('feat.zip','feat.nidm.zip')
         destpath = os.path.join(tmp_dir,destname)
         nidm_zip = zipfile.ZipFile(destpath, 'w', zipfile.ZIP_DEFLATED)
         rootlen = len(feat_dir) + 1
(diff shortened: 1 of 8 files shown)


neurovault/apps/statmaps/utils.py
```diff
@@ -281,16 +283,103 @@ def populate_nidm_results(request,collection):
...
+
+
+
+
+
+
+
+
+
+
```
chrisfilo: ?

chrisfilo:
Excellent. Please use the data I've sent you for test, add dependencies to requirements.txt and update puppet.

neurovault/apps/statmaps/views.py
```diff
@@ -350,6 +352,11 @@ def upload_folder(request, collection_cid):
                             request.FILES['file'] = f
                             populate_nidm_results(request,collection)
                             continue
+                        elif fnmatch(f.name,'*.feat.zip'):
```
chrisfilo: This is not how I imaging people will use it. FEAT directories will be given to us either through "Upload directory" mechanism or "Upload archive" (where we should not expect special .feat.zip extension). Thus we need to try to detect if the uploaded folder structure is FEAT or not.

infocortex: OK, that should be doable- the directory structure of the uploaded folder is lost, but it can be reconstructed with the `paths[]` post variable.

One thing that might need refinement, is a specification for conclusively detecting the directory types.  I can start with some reasonable assumptions for that, and I'm sure we can clarify it later if necessary.

neurovault/apps/statmaps/utils.py
infocortex: just some accident whitespace in the commit.

infocortex:
Hi, I made a test script [check_feat_dirs.py](https://github.com/infocortex/NeuroVault/blob/b303b70c853317ed1dbaa8ddba671b2f65281c3c/scripts/check_feat_dirs.py) to test the nidm-results_fsl exporter against the whole large test dataset chris shared.

At the moment, nidm-results_fsl isn't successfully parsing the data, I'm seeing this error for each subject:
```python
found feat directory at '/vagrant/FEAT/ds116/group/model001/task001/cope001.gfeat/cope1.feat'
Unable to parse the FEAT directory:
  File "/opt/nv_env/NeuroVault/scripts/check_feat_dirs.py", line 23, in <module>
    fslnidm.parse()
  File "/opt/nv_env/lib/nidm-results_fsl/fsl_exporter/fsl_exporter.py", line 66, in parse
    self._get_coordinate_system()
  File "/opt/nv_env/lib/nidm-results_fsl/fsl_exporter/fsl_exporter.py", line 509, in _get_coordinate_system
    custom_space = (self._search_in_fsf(custom_re) == "1")
  File "/opt/nv_env/lib/nidm-results_fsl/fsl_exporter/fsl_exporter.py", line 540, in _search_in_fsf
    info = info_found.group('info')
AttributeError: 'NoneType' object has no attribute 'group'
```

@cmaumet , perhaps you might have an idea on what is amiss? 

(this is on the `convert_to_turtle` branch that works with the original example.)

chrisfilo:
It seems that there is something "missing" in the .fsf file. I'm afraid
@cmaumet is on holiday. Could you try to figure out which line the
converter is missing in the new fsf file? A good starting point would be to
add a more meaningful exception (that include that filename and the regexp)
here:
https://github.com/cmaumet/nidm-results_fsl/blob/convert_to_turtle/fsl_exporter/fsl_exporter.py#L538

infocortex:
I was able to get it working with a few small changes: https://github.com/infocortex/nidm-results_fsl/commit/e332ddc9002dcb5f2927df73d59031c94080c9f5?w=1

- alternateReference_yn setting was missing in design.fsf
- deleting calculated_sigmasquareds.nii.gz was failing because it seems to be needed again later
- directories such as cope001 with leading zeros was a problem in the contrast number parsing

So, we should be set- all though I guess we might see more variations in the format parsing with other datasets.

chrisfilo:
Excellent - it would be great if you could send them a PR.

infocortex:
OK, I think we're all ready to deploy:

- Unit tests: fed4fa6
- Added a modal processing dialog during upload, since the FEAT directories take quite some time: 0a05e9c
- Puppet installation of FEAT libaries: https://github.com/NeuroVault/neurovault_puppet/pull/15
- Modified changes to `nidm-results_fsl` exporter (I'll PR shortly): https://github.com/infocortex/nidm-results_fsl/compare/nv_testdata_fixes

infocortex:
PR sent for nidm-results_fsl: https://github.com/cmaumet/nidm-results_fsl/pull/1.

chrisfilo:
Despite the fact that test pass when I try to upload any of the example data .zip files (from openfmri or the one from Camille) I get redirected to /collections/13/upload_folder. What is more when I go back to the collection there is no sign of any new NIDM Results records. Can you replicate this?

chrisfilo:
It also needs to work when "Upload folder" (uncompressed) is being used.

infocortex:
Sorry, that was recently introduced, fixed in  0a05e9c.

chrisfilo:
After those fixes I get redirected back to the collection page (which is good), but no NIDM files are added (when I try with the two openfmri examples). When I use Camilles example it is treated like any other .zip folder (in other words all nifti files are added to the collection). Could you improve the tests or try those three files in order to find and fix the bug?

infocortex:
These are all working for me, as zips and folders.

By the openfmri examples, you mean are the ones already added as the test files?

Not sure what's going wrong.  Can you send me the failing zips?

infocortex:
Could your javascript still be cached, perhaps?

chrisfilo:
I tried with fresh vagrant instance and cleared cache (using a plugin as well as icognito mode). This is the behaviour I observe:

- for ds017.zip and ds005.feat.zip from the test_date folder I get redirected to the collection page after the upload with no new items in the collection
- if I upload corresponding folders instead of .zip files for those two datasets the behaviour is the same but I get a "Processing upload" dialog box. Still no new items in the collection when it's done processing
- if I use Camilles example ex_fmri_one_contrast.feat.zip nifti files are added to a collection as if it was a normal zip file. No new NIDM records.

infocortex:
OK, these issues should be resolved.
- uwsgi issue fixed in  2405685 and https://github.com/NeuroVault/neurovault_puppet/commit/e3b9190dfbd7edec9d4a261752939e4aaf1d494a
- I also added a error notification for failures in folder and zip uploads: 5c40704

infocortex:
True in, but it shouldn't be a problem with the path being explicitly added in our app.

Also, if the provconvert binary fails for some other reason, the subsequent exception in the nidm parsing will now throw a visible exception.  (nidm.ttl won't exist in a usable state).

chrisfilo:
I just tested it - without fixing the path it still fails silently (instead of rising an exception). After fixing the path it works (for openfmri examples). It still fails to recognise Camille's example as FEAT dir.

chrisfilo:
I've updated nidmresults to raise an exception if the conversion fails https://github.com/incf-nidash/nidmresults/pull/2 . To test it I on purpose removed provconvert from path and now I am getting this error:

	ValueError at /collections/5/upload_folder
	The view neurovault.apps.statmaps.views.upload_folder didn't return an HttpResponse object. It returned None instead.

infocortex:
Sounds good, catching at the subprocess call is definitely preferable.

The related error handling is fixed in  8225839.  I tested normal error cases, and also a missing provconvert binary, for zips and folders.

> It still fails to recognise Camille's example as FEAT dir.
Make sure you aren't still using the original copy of that file that had no contents in the logs/ or stats/ directory, that is likely the issue.

PR #35, opened 2014-11-05
ENH: check if map is unthresholded

jbwexler:

PR #111, opened 2015-01-24
Enh/image comparison

vsoch:
This will add image comparison to NeuroVault, meaning:  a button on image pages that will go to a search interface to show top 100 most similar images, each of  which is linked to the "scatterplot compare" page for the two images.  The query image is in the top right, and image types (Z, beta, Other, etc) are in the top left, so the user can click to filter.  From each image you can also go to the image page.  The glass brains do NOT represent a difference / comparison, but are just the maps themselves, so the user can visually see them vs. the query in the top right.

The images that are included in the comparison come directly from a pkl file that contains a similarity (pearson) correlation matrix, with row and column names corresponding to the image pks.  The button will only show up for an image if it is in this set.  Two things need to happen for this to work:

- the pkl matrix must exist
- the png images (one for each brain map) must also exist

This pull request contains the tasks.py and other celery modules required to get this working, with the tasks written, but they are not integrated anywhere into neurovault for it to happen automatically.  We would need to:

- submit job to celery to generate the brain image glass brain png with generate_glassbrain_image(nifti_file,pk) in statmaps/tasks.py when a new image is added
- (on some timely basis) run "make_correlation_df" (also in statmaps/tasks.py) to generate the public image data frame.  

The png brain glass images go into the image/# folders along with the images, with the format "glass_brain_%s.png" % pk, and the correlation matrix is a pickle that gets saved to  image_data/matrices/pearson_corr.pkl

The guts of the image comparison come from the pybraincompare package, which basically takes in images and stuffs and spits out html, which is then rendered into the neurovault templates.  Here is a list of packages I remember needing to install for my edits (probably these guys go into neurosynth_puppet?):

- [pybraincompare](https://github.com/vsoch/pybraincompare)
- scikit-learn
- nilearn

And another change for puppet would be adding the "matrices" folder under image_data/matrices 

Finally, I'm not sure how the actual deployed image paths looked - if you look under compare_images in views.py I changed the /opt/image_data bits to media and that seemed to work, and I did the same for the compare_search function. 

There are of course much cool things we can add on to this (let the user choose an atlas, more filtering options, show some kind of difference map), but since this is my first contribution I want to start small... baby steps!  I've also never done this before so feel free to be blunt with me in giving instructions! :L)

### DIFF
--- a/neurovault/apps/statmaps/views.py
+++ b/neurovault/apps/statmaps/views.py
@@ -725,6 +725,7 @@ def find_similar(request,pk):
     for comp in comparisons:
         image_ids.append([image_id for image_id in [comp.image1_id,
                          comp.image2_id] if image_id != pk][0])
+    images_processing = len(Image.objects.all()) - len(image_ids)
     scores.insert(0,pk)
     data = pandas.Series(scores,index=image_ids, name=pk)
 
@@ -755,7 +756,7 @@ def find_similar(request,pk):
                                              image_url=image_url,query=query,absolute_value=True,
                                              max_results=100,image_names=image_names)
     html = [h.strip("\n") for h in html_snippet]
-    context = {'html': html}
+    context = {'html': html,'images_processing':images_processing}
     return render(request, 'statmaps/compare_search.html', context)
 
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,5 +36,6 @@ rdflib>=4.1.0
 celery[redis]
 django-celery
 scikit-learn
+python-redis
 -e git://github.com/nilearn/nilearn.git@master#egg=nilearn
 -e git://github.com/vsoch/pybraincompare.git@955182ef6068dc056ab16586ef7c70e65d87301c#egg=pybraincompare
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,6 +36,6 @@ rdflib>=4.1.0
 celery[redis]
 django-celery
 scikit-learn
-python-redis
+django-redis
 -e git://github.com/nilearn/nilearn.git@master#egg=nilearn
 -e git://github.com/vsoch/pybraincompare.git@955182ef6068dc056ab16586ef7c70e65d87301c#egg=pybraincompare
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,6 +36,5 @@ rdflib>=4.1.0
 celery[redis]
 django-celery
 scikit-learn
-django-redis
 -e git://github.com/nilearn/nilearn.git@master#egg=nilearn
 -e git://github.com/vsoch/pybraincompare.git@955182ef6068dc056ab16586ef7c70e65d87301c#egg=pybraincompare
--- a/scripts/add_pearson_similarity_metric.py
+++ b/scripts/add_pearson_similarity_metric.py
@@ -9,7 +9,7 @@ os.environ.setdefault("DJANGO_SETTINGS_MODULE", "neurovault.settings")
 django.setup()
 
 from neurovault.apps.statmaps.models import Similarity, Comparison, Image
-from neurovault.apps.statmaps.tasks import save_voxelwise_pearson_similarity, update_voxelwise_pearson_similarity
+from neurovault.apps.statmaps.tasks import save_voxelwise_pearson_similarity
 from django.db import IntegrityError
 import errno
 
--- a/neurovault/apps/statmaps/templates/statmaps/compare_search.html
+++ b/neurovault/apps/statmaps/templates/statmaps/compare_search.html
@@ -17,6 +17,20 @@ var addthis_config = {
 
 <script type="text/javascript"> 
 
+var portfolio_filter = document.getElementsByClassName("portfolio-filter")
+var li = document.createElement("li");
+var a = document.createElement("a");
+
+function setAttributes(el, attrs) {
+  for(var key in attrs) {
+    el.setAttribute(key, attrs[key]);
+  }
+}
+
+setAttributes(a, {"class": "btn btn-warning", "disabled": "true"});
+a.appendChild(document.createTextNode(["{{ images_processing }} images still processing"]));
+portfolio_filter[0].appendChild(li.appendChild(a))
+
 $(document).ready(function() {
 
     //$('li.portfolio-item').find('a.btn-danger').click(function(event) {


vsoch:
and I think it's reasonable to start just generating the correlation matrix pickle and png images manually, and then we can figure out this celery business!

vsoch:
One more quick note! The last (and first) time I tried merging conflicts, I royally messed it up, because I don't quite get how it works with all the ################# inserted into the files. So I'd like to see how this is done properly before royally messing something up that I can't fix.

neurovault/apps/statmaps/views.py
```diff
@@ -146,10 +149,13 @@ def view_image(request, pk, collection_cid=None):
     image = get_image(pk,collection_cid,request)
     user_owns_image = True if owner_or_contrib(request,image.collection) else False
     api_cid = pk
+    # Comparison is possible if pk is in matrix columns
+    corr_df = pandas.read_pickle("/opt/image_data/matrices/pearson_corr.pkl")
```
chrisfilo: don't use abs paths

neurovault/apps/statmaps/views.py
```diff
@@ -550,6 +556,68 @@ def atlas_query_voxel(request):
...
+    image2 = get_image(pk2,None,request)
+    
+    # Name the data file based on the new volumes
+    path, name1, ext = split_filename(image1.file.url)
+    path, name2, ext = split_filename(image2.file.url)
+
+    # image1.file.url is not correct --> /media/images does not exist
+    # hardcoded for now- for Neurovault can change volume1 to image1.file.url and it should work
+    volume1 = image1.file.url.replace("/media/images","/opt/image_data/images")
+    volume2 = image2.file.url.replace("/media/images","/opt/image_data/images")
```
chrisfilo: again - no hardcoded paths

requirements.txt
```diff
@@ -31,4 +31,6 @@ requests-oauthlib
 python-openid
 django-sendfile
 django-polymorphic
-networkx
\ No newline at end of file
+networkx
+celery-with-redis
+django-celery
```
chrisfilo: you should add other dependencies here

chrisfilo:
Don't worry about messing up when learning how to resolve conflicts. There is no way you can screw anything up. Just don't do "push --force" and you will not risk any data loss. Here are some good pointers on how to deal with conflicts: https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/

neurovault/apps/statmaps/views.py
infocortex: you need `image.file.path`, `image.file.url` is a virtual url used to obfuscate the real path

neurovault/apps/statmaps/tasks.py
```diff
@@ -0,0 +1,60 @@
...
+  # Get all image paths
+  image_paths = [image.file.path for image in public_images]
+  images_resamp, reference_resamp = resample_multi_images_ref(image_paths,reference,resample_dim)
+  # Read into pandas data frame
+  corr_df = pd.DataFrame()
+  row_count = 0
+  for image in images_resamp:
+    corr_df[row_count] = image.get_data().flatten()
+    row_count = row_count + 1
+  corr_df = corr_df.corr(method="pearson")
```
chrisfilo: shouldn't we create an implicit mask that would consist of an intersection of nonzero voxels from both images before calculating the correlation?

infocortex:
Regarding celery, theres two main use cases with celery that I'll be providing example implementations we can base off of:

1) The client instantiates a background task and waits for it to complete in a single view, with a progress bar for iterative and predictable tasks, or  a waiting animation.  (This would apply to to the  comparison search)

2) A background task is triggered with a save event, based on a edit or save of a database object, (as you were working on with the glass brain generation, which I think didn't work because it needs to be in a `post_save()` method, I'll give that a shot first.)

For the (re)generation of the comparison matrix, I'm assuming we also need option 2, as it needs to be modified any time an image is uploaded, changed or deleted, unless I'm mistaken.

So, more to come on the Celery side of things, but thats the basic gist of it.  

PS - Glass brains are neat.

infocortex:
Ah, actually I see now we want to run the matrix generation as a periodic task, rather than at save/edit time.

vsoch:
awesome thanks for the comments and feedback! I'll be on these changes first thing tomorrow morning :O)

vsoch:
yes! I'm not sure if celery can do more "cron job" sort of things versus "save png brain" at edit time.

neurovault/apps/statmaps/tasks.py
vsoch: Wouldn't that mean, given a vector of pearson correlations for image A against B, C, D.. Z - we wouldn't be comparing apples and oranges? For example, in an extreme case, image D has one non zero voxel. If we do an intersection, then we are comparing just that one voxel, and the similarity score could be ridiculously high. Our search would be biased to return more sparse maps, because we can't really control if users input thresholded maps.  

I think what you are hinting at (and the ideal) is to have a smarter comparison strategy that takes into account the intersection, size, etc, and I am in agreement, but I haven't really started to think about this yet (it goes well with aim 1!)  If the implicit mask is considered a standard and you say "Vanessa, just do it," then I definitely will go for it! But if people are sort of ambivalent about it, maybe we should do what is silly and straightforward... a correlation from all voxels within the same brain mask? Let me know your thoughts.

vsoch:
I just tried making a new branch and resolving conflicts... it resulted in an internal server error (no other information in the browser) so... I don't really know how to do this right.

The nginx log says:

2015/01/23 17:24:40 [error] 1446#0: *383 upstream prematurely closed connection while reading response header from upstream, client: 192.168.33.1, server: localhost, request: "GET /images/3/ HTTP/1.1", upstream: "uwsgi://unix:/tmp/neurovault.sock:", host: "192.168.33.10"

but that is from a while ago!

The last in the uwsgi log says:

Sat Jan 24 07:28:09 2015 - announcing my loyalty to the Emperor...

...

.....

....... (slightly more dots)

I think this was the straw that broke the camel's back... uwsgi and my computer are plotting against me. I really should have gotten him a new keyboard last month.

vsoch:
The above was an attempt at a joke :P I'll give this a look with fresh eyes tomorrow and hopefully figure it out :)

vsoch:
ok, I think I did it! When I was looking at the merge stuffs, I saw lots of NIDM additions, and then in the requirements.txt I realized I didn't have rdflib, and so installing that got the django error messages popping up again.  Then I saw some error about a statmap_image.figure, and I saw that Chris added a new field to that database https://github.com/NeuroVault/NeuroVault/commit/99ad2428457aa6b4dde7382d3d7d68c833404931 and you can automatically update the database with those "migration" files (manage.py migrate) and then the error went away!

I (think) I addressed the comments, but let me know if I missed something - I was working on it last night without contact lenses which is highly dangerous.

neurovault/apps/statmaps/tasks.py
```diff
@@ -0,0 +1,60 @@
...
+def generate_glassbrain_image(nifti_file,pk):
+  nifti_file = str(nifti_file)
+  png_img_name = "glass_brain_%s.png" % pk
+  png_img_path = os.path.join(os.path.split(nifti_file)[0],png_img_name)
+  glass_brain = plot_glass_brain(nifti_file)
+  glass_brain.savefig(png_img_path)
+
+
+@shared_task
+def make_correlation_df(resample_dim=[4,4,4],pkl_path="/opt/image_data/matrices/pearson_corr.pkl"):
```
chrisfilo: remove absolute paths

neurovault/apps/statmaps/tasks.py
chrisfilo: Both methods have biases, and there is not standard. We need to investigate this more, let's leave it as it is for now.

neurovault/apps/statmaps/templates/statmaps/image_details.html.haml
```diff
@@ -0,0 +1,83 @@
...
+            - if image.tags.all
+	            %p <b>Tags</b>:
+	                - for tag in image.tags.all
+	                    %a{ href: "/images/tags/{{tag.name}}" }= tag.name
+	                    - if not forloop.last
+            %a.btn.btn-primary.btn-large{href: "pycortex" } 3D View
+            %a.btn.btn-primary.btn-large{href: "{{ image.file.url }}" } Download
+            %a.btn.btn-primary.btn-large{href: "http://neurosynth.org/decode/?neurovault={{ api_cid }}" } Decode with neurosynth
+            - if comparison_is_possible
+                %a.btn.btn-primary.btn-large{href: "/search/{{ image.id }}" } Find similar
```
chrisfilo: use "url" syntax instead of hard coded paths

neurovault/apps/statmaps/templates/statmaps/statisticmap_details.html.haml
```diff
@@ -80,11 +80,13 @@
             - else
                 %a.btn.btn-primary.btn-large{href: "{{ image.file.url }}" } Download
             %a.btn.btn-primary.btn-large{href: "http://neurosynth.org/decode/?neurovault={{ api_cid }}" } Decode with neurosynth
+            - if comparison_is_possible
+                %a.btn.btn-primary.btn-large{href: "/search/{{ image.id }}" } Find similar
```
chrisfilo: use url syntax

neurovault/apps/statmaps/urls.py
```diff
@@ -137,6 +137,14 @@ def get_queryset(self):
...
     url(r'^api/atlas_query_voxel/$',
         atlas_query_voxel,
-        name = 'atlas_query_voxel')
+        name = 'atlas_query_voxel'),
+
+   # Compare images
+    url(r'^compare/(?P<pk1>\d+)/(?P<pk2>\d+)$',
+        compare_images,
+        name='compare_images'),
+    url(r'^search/(?P<pk1>\d+)$',
```
chrisfilo: could we change this to r'^(?P<pk1>\d+)\find_similar$' ?

neurovault/apps/statmaps/views.py
```diff
@@ -650,6 +658,63 @@ def atlas_query_voxel(request):
...
+    image1 = get_image(pk1,None,request)
+    # This df should ONLY contain public images! see tasks.py for generation
+    corr_df = pandas.read_pickle(os.path.abspath(os.path.join(os.path.dirname( neurovault.settings.BASE_DIR ), '../..', 'image_data/matrices/pearson_corr.pkl')))
+    public_images = Image.objects.filter(collection__private=False,id__in=corr_df.columns)
+
+    # Get all image png paths
+    image_paths = [image.file.path for image in public_images]
+    png_img_names = ["glass_brain_%s.png" % image.pk for image in public_images]  
+    png_img_paths = [ os.path.join(os.path.split(image_paths[i])[0],png_img_names[i]) for i in range(0,len(image_paths))]
+    png_img_paths = [imagepath.replace("opt/image_data","media") for imagepath in png_img_paths]
```
chrisfilo: you cannot assume that opt/image_data will be there - this depends on the installation details. @infocortex what's a better way of doing this?

requirements.txt
```diff
@@ -33,3 +33,8 @@ django-sendfile
 django-polymorphic
 networkx
 rdflib>=4.1.0
+celery-with-redis
+django-celery
+scikit-learn
+nilearn
+pybraincompare
```
chrisfilo: I don't think those are in pypi so please add git link (see http://stackoverflow.com/questions/16584552/how-to-state-in-requirements-txt-a-direct-github-source)

chrisfilo:
Where do you use "image_details.html.haml"? Wasn't it removed in the past?

neurovault/apps/statmaps/views.py
```diff
@@ -650,6 +658,63 @@ def atlas_query_voxel(request):
...
+    compare_url = "/compare" # format will be prefix/[query_id]/[other_id]
+    image_url = "/images" # format will be prefix/[other_id]
+    
+    # Here is the query image
+    query = os.path.join(os.path.split(image1.file.path)[0],"glass_brain_%s.png" %(image1.pk))
+    query = query.replace("opt/image_data","media") # This may need to change for deployment?
+    
+    # Do similarity search and return html to put in page
+    html_snippet = compare.similarity_search(corr_df=corr_df,button_url=compare_url,image_url=image_url,query=query)
+    html_snippet = html_snippet[9:len(html_snippet)-2]  # get rid of body and html tags
```
chrisfilo: this is super hacky - maybe there should be an option in your toolbox to return something embeddable instead of full website?

neurovault/apps/statmaps/views.py
```diff
@@ -650,6 +658,63 @@ def atlas_query_voxel(request):
...
+def compare_images(request,pk1,pk2):
+    image1 = get_image(pk1,None,request)
+    image2 = get_image(pk2,None,request)
+    
+    # Name the data file based on the new volumes
+    path, name1, ext = split_filename(image1.file.url)
+    path, name2, ext = split_filename(image2.file.url)
+
+    # For now, manually choose an atlas - this can be user choice
+    atlas_file = os.path.abspath(os.path.join(os.path.dirname( neurovault.settings.BASE_DIR ), '../..','image_data/atlas/MNI-maxprob-thr25-2mm.nii'))
```
chrisfilo: this should be added to static and commited to the repo instead of "image_data" - like we did with the MNI template

neurovault/apps/statmaps/tests/test_celery.py
```diff
@@ -0,0 +1,31 @@
+# in one terminal window:
```
chrisfilo: how difficult would it be to turn this into a real django test that does not need to be run automatically? https://docs.djangoproject.com/en/1.7/topics/testing/ http://docs.celeryproject.org/en/2.5/django/unit-testing.html

neurovault/apps/statmaps/views.py
```diff
@@ -165,6 +167,11 @@ def view_image(request, pk, collection_cid=None):
     user_owns_image = True if owner_or_contrib(request,image.collection) else False
     api_cid = pk
 
+    # Comparison is possible if pk is in matrix columns
+    corr_df = pandas.read_pickle(os.path.abspath(os.path.join(os.path.dirname( neurovault.settings.BASE_DIR ), '../..', 'image_data/matrices/pearson_corr.pkl')))
```
chrisfilo: this will still fail if the private media path in settings changes (for example in a different installation) - better use the variable from settings

neurovault/apps/statmaps/tests/test_celery.py
vsoch: I'm not sure about the image_details.html.haml, I don't think that I use it... should I delete?

chrisfilo: yes - probably this happened when you resolved conflicts

neurovault/apps/statmaps/views.py
chrisfilo: When I think more about this it's weird. Nothing should be in media AFAIK.

vsoch: Oh, I think I fixed this one, I'm not sure why it is the old one up here! Take a look here: https://github.com/NeuroVault/NeuroVault/pull/111/files#diff-7129b607d02d4a1b46380abd3fd7d834R692 it was the same suggestion by Gabriel to use the image.file.url instead. I'm not totally sure about how this works, but (I think) the media/images path isn't actually a folder but just a URL path that goes to wherever image stuffs is: https://github.com/vsoch/NeuroVault/blob/master/neurovault/apps/statmaps/urls.py#L105

requirements.txt
```diff
@@ -33,3 +33,8 @@ django-sendfile
 django-polymorphic
 networkx
 rdflib>=4.1.0
+celery-with-redis
+django-celery
+scikit-learn
+nilearn
+-e git://github.com/vsoch/pybraincompare.git@fe0efe78e23ffb44f709230198a442cb117c7f0a
```
chrisfilo: do we need that specific commit or are you plannig to provide a backwards compatible API?

neurovault/apps/statmaps/views.py
```diff
@@ -650,6 +658,60 @@ def atlas_query_voxel(request):
...
+    corr_df["png"] = png_img_paths
+    corr_df["tags"] = tags
+    compare_url = "/compare" # format will be prefix/[query_id]/[other_id]
+    image_url = "/images" # format will be prefix/[other_id]
+    
+    # Here is the query image
+    query = os.path.join(os.path.split(image1.file.url)[0],"glass_brain_%s.png" %(image1.pk))
+    
+    # Do similarity search and return html to put in page
+    html_snippet = compare.similarity_search(corr_df=corr_df,button_url=compare_url,image_url=image_url,query=query)
```
chrisfilo: where in code can I control how many similar images are displayed? Are you including high negative correlations as well?

vsoch:
@chrisfilo, the number of max results to return definitely needed to be a variable, so I added that. The inability to specify taking the absolute value was a very good catch, and a very bad bug! @infocortex atlas static files have been moved to where they should be!

chrisfilo:
tests fail

	......E
	======================================================================
	ERROR: neurovault.apps.statmaps.tests.test_celery (unittest.loader.ModuleImportFailure)
	----------------------------------------------------------------------
	ImportError: Failed to import test module: neurovault.apps.statmaps.tests.test_celery
	  File "/usr/lib/python2.7/unittest/loader.py", line 254, in _find_tests
	    module = self._get_module_from_name(name)
	  File "/usr/lib/python2.7/unittest/loader.py", line 232, in _get_module_from_name
	    __import__(name)
	  File "/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tests/test_celery.py", line 18, in <module>
	    class Test_celery_tasks(pk):
	NameError: name 'pk' is not defined

chrisfilo:
"generate_glassbrain_image" is only called in tests - shouldn't it be called somewhere by celery?

vsoch:
yeah I didn't actually test the tests... :O)

vsoch:
it should, but when I import the task into the models.py where the image is created I get a server error, and no way to debug. I figured I had celery setup incorrectly and was going to wait for Gabriel's feedback.

neurovault/apps/statmaps/tasks.py
```diff
@@ -0,0 +1,63 @@
...
+  glass_brain = plot_glass_brain(nifti_file)
+  glass_brain.savefig(png_img_path)
+
+
+@shared_task
+def make_correlation_df(resample_dim=[4,4,4],pkl_path=None):
+  if not pkl_path:
+    pkl_path = os.path.abspath(os.path.join(neurovault.settings.PRIVATE_MEDIA_ROOT,'matrices/pearson_corr.pkl'))
+  # Get standard space brain
+  reference = os.path.join(os.environ['FREESURFER_HOME'],'subjects', 'fsaverage', 'mri', 'brainmask.nii.gz')
```
chrisfilo: This file does not exists - shouldn't this be brain.nii.gz like in https://github.com/NeuroVault/NeuroVault/blob/683763f6c5e8a2d80e94ded7761405614feb87c7/neurovault/apps/statmaps/utils.py#L92

chrisfilo:
I managed to get it to run today. Here are some comments:

1. most map names don't make much sense without the name of the collection they belong to. IT would be great if we could have names of maps and collections (maybe somehow truncated) in both "find_similar" and "compare" views.
2. There should be a way to go from compare view to individual maps
3. Correlation values in the compare view are hard to read for deselected regions
4. I wish there was an easy way to deselect all regions apart from one in compare view
5. I'm not sure we should download js from external servers (d3 for example) - what do you think @infocortex ?

Otherwise - great job!

vsoch:
1. Map and collection names have been added to both "find similar" and "compare" views
2. Compare view links to individual maps by clicking on axis titles (the image names, this is what seemed most intuitive to do)
3. Compare values turn black when the region is de-selected (and gray)
4. There is now a "clear" button that will deselect all regions :)
5. It's my opinion that getting web resources from CDNs is ideal, and given how the pybraincompare module "packages" all rendered html into a nice snippet, the external link is an ideal strategy for that.

chrisfilo:
Excellent!

In terms of performance and memory. The more I think about it the more I come to a conclusion that we should store similarity measures in the database - one entry per pair of maps. This way when a map gets updated we could delete all the similarity measures between this map and all other maps and spawn celery jobs to recalculate them. When generating "find_similar" page we would not have to load a pkl file (which could've been locked because some celery job was writing to it at the moment) and just query the database (much faster). What do you think?

vsoch:
Hmm, that's a good idea. Let's think about this - NeuroVault has SQL on the back end? That would translate to adding a new column and row each time there is a new image, and updating specific indices in that matrix when a map is updated.  From a database standpoint, it makes sense because we can have many operations going on at once (vs the trouble with any sort of "flat file" solution).  On the other hand, it would get complicated when we extend image comparison across more than one metric, or different transformations of the data, or some parcelation. I think we are reaching a point where there are two types of NeuroVault data:

- analysis related that changes frequently, is HUGE, and less "valuable"
- core app data that changes less frequently, is reasonably sized, and very valuable 

So the two questions on my mind:

- should we put those two things in the same place?
- is there a better solution to handle the operations of doing analysis that could handle the temporality and huge size?

This actually might be a really good application for bigquery - I've used it just a little bit and it seems to handle large data well.  I would want to first read about how well it integrates with django apps [there is stuff out there!](https://github.com/tylertreat/BigQuery-Python), and then could do some testing to see the feasibility of doing the computation!  [This would likely be the deal breaker](https://cloud.google.com/bigquery/pricing), but it seems dirt cheap, and my experience says that we could probably approach google and get some initial support!

vsoch:
I do think a database solution for the similarity (other future analysis stuffs) is the right way to go!

chrisfilo:
Adding and removing columns to a table is not a good idea. I much prefer a much simpler, but easier to query solution when each pair is one row. You can index by ids of the two images and have as many similarity metrics as you want. The model would look like this:

	class Similarity(Model):
		image1 = ForeignKey(Image)
		image2 = ForeignKey(Image)
		similarity_metric1 = FloatField()
		
		class Meta:
	        	unique_together = (("image1", "image2"),)
	        	
You of course would have to be smart about populating and querying this relation so image1/image2 similarity would be always image2/image1 similarity without duplicating data, but I'm sure this is doable.

I don't think we need BigQuery at this stage

vsoch:
Ah, ok, I understand. Should I open a new branch to give this a go, or add to the current enh/image_comparison?

chrisfilo:
You can create a new branch if you prefer working this way (experimenting with different database setups), but eventually I would like to see the final changes in this PR (instead of a new one). So if you create a new branch you would have to merge it into this one an push to github so the changes would show up here.

vsoch:
ok - I'm going to use a different one just in case I mess things up, and it would ultimately wind up here. I assume the goal would be to get this done for this next version of NeuroVault? What is the deadline?

tyarkoni:
Not sure about going with a DB solution... In the short to medium term that will work fine, but the size will grow quadratically, and even with 10,000 maps (not too far off!) you'll already have 50 million rows. It will also become progressively more time-consuming to compute the correlations to put in the database with every image (though that will probably take a much longer time to become an issue).

A better way to go might be to store a dimensionally-reduced version of each image in memory. I would guess that you could reduce each image from ~200k voxels to something like 500 bins (e.g., by averaging within all the voxels in each cluster in a parcellation) with relatively little loss of fidelity. The absolute r values might change some, but the rank-ordering should remain fairly stable. And I don't think it matters much if the correlations returned are an approximation, seeing as there's probably a considerable amount of error in almost all of the images. You could probably handle a couple hundred thousand images easily without having to worry about memory issues (e.g., 100k images would require ~200MB of memory at single precision).

At some point, it might also make sense to do some clustering/culling of images. I would guess that a lot of these images are very similar to one another, especially within-collection. It's probably not terribly useful to be told, for instance, that the 10 most similar images to the target are all from the same collection. I can image a simple heuristic that only retains multiple images from a given collection if they correlate below, say, .95 with every other image in the collection. Conceptually this is similar to what I do with Neurosynth, where I only store a subset of the available features in memory for the similarity analysis, because otherwise there would be a lot of redundancy in the ranking.

Down the line if things get really big, you could also take a look at something like the asymmetric hashing Google does for real-time Google Correlate searches, which is pretty clever (http://www.google.com/trends/correlate/nnsearch.pdf).

vsoch:
Thanks Tal. I'm definitely in agreement about down the line needing some kind of clustering, or minimally a strategy for deciding which similarities are most meaningful to store.  The extreme / obvious case is that an image with 0% similarity isn't going to make sense to remember.  If we needed that set, we would just take the inverse of the scores we did keep, for example.

and we will need to properly determine the amount of information that is lost (in similarity metrics) at different transformations. That's actually going to be quite fun! And heeey now... this is starting to sound a lot like a thesis project... :O)

vsoch:
and this looks like an awesome paper! Thanks!

chrisfilo:
When it get's too big to store we can always keep say 100 top correlations
in the database and recalculate everything else only when requested.

vsoch:
The update above includes models to handle image comparison, including:

- Similarity: to hold a similarity metric and a transformation of the data
- Comparison: to hold two images, and the comparison between them

that were created with migrations (I hope I did that right!) and scripts to calculate comparisons:

- with a celery task https://github.com/vsoch/NeuroVault/commit/d9842ccf4f22d2eb9fa46bf9f924cbeb8c763c9a#diff-1ba40f24f9d683d4250e2176540ffc9cR53
- and [over the current database](https://github.com/vsoch/NeuroVault/blob/d9842ccf4f22d2eb9fa46bf9f924cbeb8c763c9a/scripts/add_pearson_similarity_metric.py)

I have not yet made changes to use this data in the actual interface - I am going to generate a new function for pybraincompare that generates an interface from a precomputed list (instead of needing to calculate the similarity / sort from a pandas matrix, as is done now).

vsoch:
I tried that first,  but putting into the data frame handles the nans really nicely, and also makes any calculations that come out equivalent to if they were done with pandas (which is a more likely way to do it than comparing 1:1)

chrisfilo:
Doesn't pandas use a intersection of not-a-nan masks when calculating correlation? Imagine you have two pairs of maps: one pair encodes out of mask voxels with zeros and other with NaNs. With current implementation the similarity will be calculated in a different way.

vsoch:
Hmm. So we should set nan's to 0 no matter what, then use pearsonr? I was more comfortable not including nan, because I'm not confident that in every single situation, a NaN is indicative of a value of 0.  If there was not data for a voxel (within the brain mask) because of dropout, for example, and then you have some Z score map with NaN, changing it to 0 would be representing it as the mean. If we use intersection of not-a-nan masks when calculating correlation, that minimally means that the way a single image is compared to the rest is consistent (in not using that voxel) and isn't that what we want? On the other hand, it does mean that when we say "voxel-wise" we really mean "the set of voxels that are not-nan for the query image" and that could then mean different sets of voxels for different comparison sets, based on the nans.  

If the goal is to, for one image, find images that are most similar to it, I'd say we might care more about only including voxels for that image we have confidence were measured (the first) and less about the fact that the set of voxels is different. What do you think?

chrisfilo:
Zeros and NaN are equivalent in statistical maps and they mean that the voxel was not inside the mask. Scanner related signal dropouts are not encoded as NaNs. Only in rare cases (NIDM Results) we will have access to the mask directly. As Tom explained in his email treating zeros as real zeros can lead to all sorts of problems. I would recommend creating a non-zero and non-nan mask for each image and using their intersection to calculate correlation. In the future we can work on similarity metric that would appropriately penalise small masks (ROIs).

vsoch:
ok, that sounds reasonable.

tyarkoni:
I would be *very* wary of using only the intersection of non-zero voxels to compute the correlation between two maps without having a small-volume penalty of some kind. What's going to happen is that the variance of correlations is going to be huge for very small regions, which means that it's very likely that once you have a couple of thousand images, the top N correlations with a whole-brain map will be dominated by small masks. This will lead people to make exactly the wrong kind of inference, and will actually reward them for uploading maps with less information.

I think a good compromise solution would be to only exclude voxels where *both* maps have a zero or NaN value. I.e., instead of first computing a mask of all the finite, non-zero values in each map and then taking the intersection of the two, sum the two images together first and then keep all the finite non-zero values.

Actually, even better would be to report the correlation both ways, because that would force the user to acknowledge the potential discrepancy in values. So maybe if it's not too expensive computationally, run the correlation both ways? If you do it this way, I'll change the way Neurosynth reports correlations from the decoder to do the same thing...

chrisfilo:
I acknowledge and appreciate your concern about users using thresholded maps. We are trying to address this problem at its root - @jbwexler is working on upload check that gives user a warning if they try to upload a thresholded map (see https://github.com/NeuroVault/NeuroVault/pull/35). To further help with this problem we should consider disabling the map comparison (and decoding) feature for thresholded maps (with a warning explaining why we do this, encouraging users to upload unthresholded versions).

Replacing missing values with zeroes (or any other arbitrary number) is not a good solution to this problem and leads to different sort of biases. It actually increases the estimation error of the correlation coefficient, by creating a systematic bias. Pairwise deletion (what I am proposing) is not the only way of dealing with missing values and has it's own problems (like variable estimation errors across comparisons), but @vsoch has a project on evaluating these metrics and soon we should know more. For now I recommend pairwise deletion (calculating correlation only on voxels that are in both maps) and disabling comparison for thresholded maps (until we will figure out how to best communicate the estimation error).

tyarkoni:
Cool, disabling comparisons for thresholded maps (both in NeuroVault and Neurosynth) sounds like a good short-term fix.

nicholst:
I just want to second @chrisfilo's concern about including "structural" zeros in a correlation... it's bound to lead for trouble.  About thresholded images, they are real trouble... for variations in field-of-view you can (almost) fairly call voxels "missing at random", but thresholded images have highly informative missingness and will leads to all sorts of (actually quite interesting) methodological problems.

vsoch:
Image comparison has been integrated with the new Comparison and Similarity modules, meaning that we no longer load similarity scores from a pickle file, but from the database itself. A few notes:

- I changed [pybraincompare](https://github.com/vsoch/pybraincompare/commit/955182ef6068dc056ab16586ef7c70e65d87301c) to work with a single series instead of a data frame. This speeds things up a bit, and gives the user more freedom to "roll their own" list of scores, instead of needing to provide an entire data frame. 
- The calculation is done via [pairwise deletion](https://github.com/vsoch/NeuroVault/blob/enh/image_comparison/neurovault/apps/statmaps/tasks.py#L101). As we've discussed, there is work to be done to determine optional image comparison given a motley crew of input maps, and I'm going to start testing this. I am on the same page that in the meantime, we should not release this, and should work toward having a "this image is thresholded!" upload warning, along with disabling comparison for thresholded maps.
- I created specific functions that will be celery jobs to [save](https://github.com/vsoch/NeuroVault/blob/enh/image_comparison/neurovault/apps/statmaps/tasks.py#L53) and [update](https://github.com/vsoch/NeuroVault/blob/enh/image_comparison/neurovault/apps/statmaps/tasks.py#L73) scores.
- Once celery is integrated to call the functions above, I'd like to do more rigorous testing of said functions. I was testing [here](https://github.com/vsoch/NeuroVault/blob/enh/image_comparison/scripts/add_pearson_similarity_metric.py) and started an "official" tests/test_comparison.py script (not yet included!)

I am going to start on the second point (testing) asap! I'll likely make an ipython notebook and will definitely share :cake: and likely I'll hold off on small changes to this code base until I've finished that.

vsoch:
Ok, I've started just some [early thinking](http://nbviewer.ipython.org/github/vsoch/brainmeta/blob/master/notebook/image_comparison.ipynb) about the problem, and sketched out draft experiments that I could do to test. I am going to spend some time reading and learning about the metrics, and then I will move into experiment 1 (or more likely some modified version of it), and then it will be easier to get feedback because I will have actual analysis / plots to share with everyone. Thanks for all the discussion @tyarkoni and @chrisfilo and for the reading @nicholst, that's where I am going right now! :smile:

scripts/add_pearson_similarity_metric.py
```diff
@@ -0,0 +1,53 @@
...
+
+# After this we should have ((N*N)-N)/2
+# N^2 is N combinations total for N objects
+# we subtract N to account for diagonal
+# we divide by 2 to take order into account
+
+M = len(Comparison.objects.all())
+N = len(Image.objects.filter(collection__private=False))
+we_should_have = ((N*N)-N)/2
+print "We have %s comparisons after generating all of them, we should have %s" %(M,we_should_have)
```
chrisfilo: This will fail if new maps were uploaded while this script is being run (which is likely since it will generate 500k correlations). I don't think we need this assert here.

scripts/add_pearson_similarity_metric.py
```diff
@@ -0,0 +1,53 @@
...
+
+image1 = Image.objects.filter(pk=1)[0]
+image2 = Image.objects.filter(pk=2)[0]
+
+# Update a similarity, we should still have N
```
chrisfilo: +1 for adding tests but they belong to tests folder and should be converted to Django testing framework and removed from this script

neurovault/apps/statmaps/tasks.py
```diff
@@ -0,0 +1,145 @@
...
+def generate_glassbrain_image(nifti_file,pk):
+  nifti_file = str(nifti_file)
+  png_img_name = "glass_brain_%s.png" % pk
+  png_img_path = os.path.join(os.path.split(nifti_file)[0],png_img_name)
+  glass_brain = plot_glass_brain(nifti_file)
+  glass_brain.savefig(png_img_path)
+  plt.close('all')
+
+
+@shared_task
```
chrisfilo: are you still using this?

neurovault/apps/statmaps/models.py
```diff
@@ -347,3 +351,33 @@ class Atlas(Image):
 
     class Meta:
         verbose_name_plural = "Atlases"
+
+
+class Similarity(models.Model):
```
chrisfilo: I am still not convinced we need this class in the production server. It will be a significant overhead without clear benefit. I am happy to be convinced otherwise.

tyarkoni: Not sure how Django handles associations, but it seems reasonable to keep this around as long as it's lazy loaded. If it's only used sporadically to look up the details of the similarity metric (e.g., once per results page). But it definitely shouldn't be eager-loaded every time a Comparison record is retrieved.

chrisfilo: We can query just the value without a problem (
https://docs.djangoproject.com/en/1.7/ref/models/querysets/#django.db.models.query.QuerySet.values),
but we will not be able to avoid a JOIN operation and we will be storying
potentially millions of IRI strings. The latter can be solved by storying
similarity metric and transformation types is separate tables (that would
however add extra two JOINs when querying).

vsoch: I think it's important because the universe of "kinds" of metrics and comparisons that we make is going to get large, and we need to keep it properly organized.  For example, a simple pearson could have a voxelwise brainmask version, a voxelwise pairwise deletion, some kind of regional version, etc. And we will most definitely be adding different metrics that are better suited for different image transformations and goals of the researcher.  It's important to have these separate from the actual Comparison data because at some point we will have a search page where the user is looking over his/her different options, and not only will we probably want to store more information about the metrics, but we definitely don't want to need to derive them from the monster table of comparisons themselves. For these reasons, I decided to model the two in this way.

neurovault/apps/statmaps/tasks.py
vsoch: We will need to put this function in somewhere as a celery task to generate the png, yes most definitely!

chrisfilo: I meant make_correlation_df. Sorry for the confusion

neurovault/apps/statmaps/models.py
chrisfilo: Cool - I had a second (closer) look at the implementation and I think it should be fine.

chrisfilo:
The two missing bits (unless I'm mistaken) are:

1. Deleting existing comparison values if a map gets updated (user uploads a new .nii file to an existing Image) and spawning a new comparison celery job to recreate those.
1. Spawn comparison jobs when na new map gets uploaded.

neurovault/apps/statmaps/views.py
```diff
@@ -650,6 +658,68 @@ def atlas_query_voxel(request):
...
+
+# Return search interface for one image vs rest
+def find_similar(request,pk):
+    image1 = get_image(pk,None,request)
+    pk = int(pk)    
+
+    # Get all similarity calculations for this image, and ids of other images
+    comparisons = Comparison.objects.filter(Q(image1=image1) | Q(image2=image1))    
+    scores = [comp.similarity_score for comp in comparisons]
+    image_ids = list(); image_ids.append(pk)
```
chrisfilo: why not

	image_ids = [pk]
	
Cramming multiple lines into one using a semicolon reduces code readability.

vsoch: I'm a noob programmer :) Will fix.

vsoch: The actual spawning of the jobs with celery, whether it be on upload or update, I was waiting to hear something from Gabriel. Is this something I should figure out?

Is deleting necessary for the update? If a user uploads a different file for a current map, then wouldn't the update be sufficient? However we DO need deletion if a user deletes a map entirely. I will add.

chrisfilo: @infocortex what's the status?

If an Image gets deleted all objects referring to it will also be deleted (due to ForeignKey mechanism) automatically (at least in theory).

This does not apply to update so you need to rerun comparisons if an Image gets updated with a new file.

infocortex:
Hi, I'm working the new database implementation into a merged celery branch: https://github.com/infocortex/NeuroVault/compare/vsoch:enh/image_comparison...enh/imgcompare_celery , which should be finished in a hour or two, then I will PR against @vsoch's branch.

Everything else is good to go except for Celery, correct? 

I will hook the similarity generation `save_voxelwise_pearson_similarity` into the image `save()` method, where the dataframe generation was previously.  

One little question:  the generation of the similarity metric is effectively static, and only needs to run once at deployment time, or if the correlation metric is altered, right?

vsoch:
The generation of the metric is static, yes, and only needs one run. I don't ever see adding a metric as something we would want/allow a user to do, @chrisfilo does that sound right? If/when we have a nice assortment of metrics to sample from, they too would be added manually. If a metric is altered, I think that is something that would be done manually, although I haven't worked with an actual deployed server, so I'm just guessing on that one :)  

For save_voxelwise_pearson_similarity, I thought it might go somewhere [here](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L261)?

We would also need to update_voxelwise_pearson_similarity if a map is changed. I'm not totally sure where that would go (I don't see an update function in Image in models.py)- possibly somewhere in [here?](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/forms.py)?

infocortex:
Running `add_pearson_similarity_metric.py` against the whole production dataset, I get the following:

```python
Calculating pearson similarity for images DC: Words and Pre-cashout vs. pre-pump searchlight results (unthresholded)
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
/opt/nv_env/NeuroVault/scripts/add_pearson_similarity_metric.py in <module>()
     31     if image1.pk < image2.pk:
     32       print "Calculating pearson similarity for images %s and %s" %(image1,image2)
---> 33       save_voxelwise_pearson_similarity(image1.pk,image2.pk)

/opt/nv_env/local/lib/python2.7/site-packages/celery/local.pyc in <lambda>(x, *a, **kw)
    165     __ge__ = lambda x, o: x._get_current_object() >= o
    166     __hash__ = lambda x: hash(x._get_current_object())
--> 167     __call__ = lambda x, *a, **kw: x._get_current_object()(*a, **kw)
    168     __len__ = lambda x: len(x._get_current_object())
    169     __getitem__ = lambda x, i: x._get_current_object()[i]

/opt/nv_env/local/lib/python2.7/site-packages/celery/app/task.pyc in __call__(self, *args, **kwargs)
    418             if self.__self__ is not None:
    419                 return self.run(self.__self__, *args, **kwargs)
--> 420             return self.run(*args, **kwargs)
    421         finally:
    422             self.pop_request()

/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tasks.pyc in save_voxelwise_pearson_similarity(pk1, pk2, resample_dim)
     41                                          similarity_metric=pearson_metric).exists():
     42
---> 43             pearson_score = calculate_voxelwise_pearson_similarity(image1, image2, resample_dim)
     44
     45             # create new Comparison with the voxelwise pearson similarity metric

/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tasks.pyc in calculate_voxelwise_pearson_similarity(image1, image2, resample_dim)
    101
    102     # Calculate correlation only on voxels that are in both maps (not zero, and not nan)
--> 103     binary_mask = make_binary_deletion_mask(images_resamp)
    104     image1 = images_resamp[0]
    105     image2 = images_resamp[1]

/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tasks.pyc in make_binary_deletion_mask(images)
    117     mask = numpy.zeros(images[0].shape)
    118     for image in images:
--> 119         mask[(image.get_data() != 0) * (numpy.isnan(image.get_data()) == False)] += 1
    120     mask[mask != len(images)] = 0
    121     mask[mask == len(images)] = 1

IndexError: too many indices for array
```

Aside from generally catching exceptions during generation, do we need other special handling for missing comparisons?

vsoch:
Do you have the primary keys for the images that are triggering the error? It might be because of a 4D image, which I think we should probably just skip, and I can add a check for that. If with the check it still triggers an error (meaning 3D triggers) then I'd like to also look at the images to see why. I'm sort of new to error handling, so I'll need feedback on that.

chrisfilo:
Most probably a stray 4D file. I've put conversion routines to get rid of
those, but maybe I missed something.

infocortex:
Yep, I just had to run `fix_4d.py` on my test data, and all is well...I didn't know there were a couple non-afni 4D ones lurking in there.

vsoch:
Ok, I was going to add a check - should I not do that? I was going to do:

     if not np.any([len(image.shape) for image in images_resamp])>3:
           binary_mask = make_binary_deletion_mask(images_resamp)
           image1 = images_resamp[0]; image2 = images_resamp[1]

           # Calculate correlation with voxels within mask
           return pearsonr(image1.get_data()[binary_mask==1],image2.get_data()[binary_mask==1])[0]
     else return 0

So if there is a 4D image we just return a similarity of 0, because there cannot be any similarity between a 3D and 4D. The other option is to return nothing to the save function and have it check for that.

chrisfilo:
those are fake 4D files - the fourth dimension has size 1. you should try
to squeeze them
http://docs.scipy.org/doc/numpy/reference/generated/numpy.squeeze.html

vsoch:
Ok - so we can be absolutely certain that there will not be any *real* 4D files?

infocortex:
Hi guys,

Nearly done celery-izing everything here, but had one thing to confirm.  The new method to determine if a comparison is possible, since the pickled dataframe is gone:

https://github.com/infocortex/NeuroVault/blob/8d7beb80c91a9aae24b5c8bdf0736a1ae55d859f/neurovault/apps/statmaps/views.py#L175-L177

I suppose we want to check if there are at least 100 Comparison records for that image e.g.

```python
num_comparisons = Comparison.objects.filter(Q(image1=image)||Q(image2=image)).count()
comparsion_possible = True if num_comparisons >= 100 else False
```
something like that sound reasonable?  (Side note: we need to `db_index` those columns in the comparison model.)

chrisfilo:
I would lean towards num_comparisons >= 1 and showing a "There are XX
images for which similarity is still being calculated. Please come back
later to see updated results" on the "find similar" view. XX is "number of
public images" - Comparison.objects.filter(Q(image1=image)||Q(image2=
image)).count()

+1 for setting db_index

vsoch:
I agree that we should have num_comparisons >= 1. @infocortex , if you want to add that check, I can offer to do the interface work to show the number of images available for comparison vs. in database.

infocortex:
> @infocortex if you want to add that check, I can offer to do the interface work to show the number of images available for comparison vs. in database.

sounds good, I'll fire off a PR to this PR branch that has the celery integration and other various minor refactoring and tweaks.  I haven't made any changes to the views or interface, or moved the view logic to celery tasks.  (I can do that later if needed, but the code in the views runs fast anyways.)

Before I do, I am having another issue worth chatting about- I'm running into the issue that `Atlas` images and untyped base `Image`s (from temporary collections) have no map_type:

```
AttributeError at /330/find_similar
'Atlas' object has no attribute 'map_type'
```

Should we be using the `BaseStatisticMap` object instead of the `Image` parent object everywhere, or just address the map_type issue with tags?  I guess it depends if we want to have Atlases in comparisons?

infocortex:
actually, `BaseStatisticMap` is a abstract model, so you wouldn't query it- we stil want to use `Image` model for the queries,  so its mostly a question of whether we want to filter Atlases from the queries, or use something other than map_type for the tag filtering buttons in the view.

vsoch:
It would be strange to have Atlases in comparisons. Down the road if NeuroVault ever has a substantial number of atlases and we define reasonable ways to compare them, then maybe, but I don't think they should pop up in the similarity result interface for now. That said, I think it probably makes sense for Atlases to have map_type atlas, @chrisfilo what do you think?

vsoch:
I think we should filter Atlases from the queries.

chrisfilo:
+1

infocortex:
OK, should be set, PR sent:

https://github.com/vsoch/NeuroVault/pull/5

There is some old commits in there, which should be harmless, but if you want to fast-forward your branch `git pull neurovault master` or similar, that should all go away.

I do think that the comparison graph should render in the background with celery probably, so I will add that in a separate PR later.

I'll send a PR for Puppet to add Celery reqs after :sandwich:

vsoch:
oh, sorry about that. Yes let me update enh/image_comparison!

vsoch:
@infocortex I just merged NeuroVault/NeuroVault master, and everything looks still to be running! :+1:

infocortex:
Awesome, here's the corrected pr, without all the extra stuff: https://github.com/vsoch/NeuroVault/pull/6

tyarkoni:
FYI, I just did some optimization on essentially the same problem and got some nice results. I'm trying to speed up the real-time comparison of NeuroVault images so I can decode against the full set of 3,000+ features rather than just 100 arbitrarily picked ones. I've now gotten it down to under a second. Here's what I did, which may be of use to you as well (if not for real-time analysis, then at least to speed up the correlations):

1. Initialize a mem-mapped array with size (n_maps, n_voxels), where n_maps is the number of maps to compare the target image with, and n_voxels is the number of random voxels to use.
2. Read in each NIFTI volume and standardize it, saving the standardized data in the memmap. This speeds up the correlation because it reduces it to one np.dot operation.
3. When a new image comes in, first standardize it, then get the correlations taking the dot product of the incoming vector and the entire memmapped array. This can be sped up further by reading blocks into memory and doing the operation in-memory (because memmapped operations don't benefit from the numerical optimization libraries, AFAIK).

This runs in O(n) time with respect to the size of the memmapped image array, so you can get arbitrarily good performance by randomly sampling voxels instead of using the whole brain. To give you an idea, when running on the whole brain, it takes ~10 seconds (including Python overhead) to compare a new image against ~3,500 Neurosynth features, but if you only use 20k random voxels, that drops to ~0.5 seconds without any additional optimization. And the correlations are essentially indistinguishable. For example, here are the top results when comparing the 'syllables' map against all other features across the whole brain:

stg                0.451782
gyrus stg          0.445574
auditory cortex    0.442763
listening          0.441936
auditory           0.441646
sounds             0.441217
speech             0.419344
sound              0.419249
acoustic           0.417486
syllable           0.406679
listened           0.406494
spoken             0.390991
music              0.390190
temporal gyrus     0.387975

...and if we use only 10k voxels:

gyrus stg           0.421030
stg                 0.403783
sounds              0.398734
auditory cortex     0.390975
auditory            0.389016
listening           0.384213
sound               0.375560
speech              0.369020
temporal gyrus      0.364080
listened            0.364053
acoustic            0.363448
auditory stimuli    0.359385
planum temporale    0.357650
temporale           0.357650

The rank-ordering is essentially unchanged, and the observed values are well within the variation you would expect to get given slight changes to the image (i.e., I highly doubt that the difference between a .44 and .4 correlation in the above is meaningful anyway).

I think in your case this exact approach probably wouldn't work because creating the initial memmap is the real time killer, and you don't want to do that every time an image is added. But I could see a hybrid strategy working well, where you periodically remap images to the drive (maybe once a day), and then compute correlations with the remaining (unmapped) images in memory.

chrisfilo:
Thanks for sharing @tyarkoni - this is very useful. We'll try the db solution for now and see how it performs. The interpretation of the varying rank order when subsampling voxels is interesting beyond optimisation. this could be used to asses validity of when trying to say A is more similar to B than to C. We'll definitely look into that in the future.

vsoch:
Ok, I've merged the branches, and I'll wait to push until I can test celery. I left a note on the PR page for @infocortex. We should have everything updated soon! :O)

vsoch:
Merge is complete! I noticed that the script was changed back to find the atlas file in the base "static" directory and not under /apps/statmaps/static.  Is this how it's supposed to be for the development server?

I'm working on adding the "fyi" to the user about images still processing right now.

vsoch:
button

vsoch:
I thought orange looked nice, and will attract the attention of the user. Red was a little too harsh.

infocortex:
Hi all, here's the Celery config changes for puppet: https://github.com/NeuroVault/neurovault_puppet/pull/16

I'll leave that open until someone else has a change to test it out.

Note: If you get a puppet error, manually uninstall nilearn and pybraincompare from the virtualenv (`pip uninstall ...`), and delete `/path/tovirtualenv/src/*`, puppet will subsequently track them properly from requirements.txt.

to restart the installed daemon (to get new changes):
`sudo service neurovault-tasks restart`

logs are under `/var/log/neurovault-tasks`

to test for local dev without running as daemons:
`./manage.py celery worker -A neurovault.celery -l info -E --autoreload`

:shipit:

vsoch:
I'm heading home for dinner but I should be able to give you feedback latest sometime tomorrow morning!

vsoch:
@infocortex  - I don't think I'm going to be able to help with this - trying to make the new virtual machine completely messed up my current (cached) machine, so now I can't even do dev work for NeuroVault :disappointed:

vsoch:
ok, I figured out how to fix, silly vagrant! [see here](https://github.com/NeuroVault/neurovault_puppet/pull/16)

vsoch:
@infocortex  what did you mean by deleting neurovault.sock? If I delete it does it regenerate automatically?

vsoch:
op, looks like it does come back!

requirements.txt
```diff
@@ -33,3 +33,9 @@ django-sendfile
 django-polymorphic
 networkx
 rdflib>=4.1.0
+celery[redis]
+django-celery
+scikit-learn
+python-redis
```
chrisfilo: ==> default: Notice: /Stage[main]/Main/Neurovault::Main[nvault-install]/Python::Requirements[/tmp/temp_requirements.txt]/Exec[python_requirements/tmp/temp_requirements.txt]/returns: Downloading/unpacking python-redis (from -r /tmp/temp_requirements.txt (line 39))
	==> default: Notice: /Stage[main]/Main/Neurovault::Main[nvault-install]/Python::Requirements[/tmp/temp_requirements.txt]/Exec[python_requirements/tmp/temp_requirements.txt]/returns:   Could not find any downloads that satisfy the requirement python-redis (from -r /tmp/temp_requirements.txt (line 39))

vsoch: oops, I think it should be django-redis! You can try eliminating it, and then when you test celery the error message actually specifies the module name.

infocortex:
Hi,

Redis is installed via apt, not pip- it is written in C, not Python.  The only packages needed in requirements.txt are `celery[redis]` and `django-celery`.

django-redis is an unrelated package (a caching backend), and python-redis is not a thing, so we don't need those.  :  ]

To install redis manually, you can use this apt-get commands paraphrased from the Puppet update:

```
sudo apt-get install software-properties-common
sudo add-apt-repository ppa:chris-lea/redis-server
sudo apt-get update
sudo apt-get install redis-server
```

infocortex:
Quick note re: periodic cron-like tasks, I'm not sure if we need them at the moment, but I left the necessary config commented in settings.py for now:

```python
CELERYBEAT_SCHEDULE = {
    'run_make_correlation_df': {
        'task': 'neurovault.apps.statmaps.tasks...',
        'schedule': timedelta(minutes=30),
    },
}
```

or you can manage periodic schedule in the django admin, which is pretty fancy:

```python
CELERYBEAT_SCHEDULER = 'djcelery.schedulers.DatabaseScheduler'
```

To test scheduled tasks when manually starting Celery, be sure to pass the `-B` option to start the celerybeat process.

chrisfilo:
I just created a fresh virtual machine, created a new user, added a new collection and three images in it. Since I could not see the comparison button I checked celery with "./manage.py celery worker -A neurovault.celery" I see the following error:

	[2015-02-10 04:44:31,990: ERROR/MainProcess] Task neurovault.apps.statmaps.tasks.save_voxelwise_pearson_similarity[2619bfaf-6b27-4bbe-96a2-20751eecda31] raised unexpected: DoesNotExist('Similarity matching query does not exist.',)
	  File "/opt/nv_env/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
	    R = retval = fun(*args, **kwargs)
	  File "/opt/nv_env/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
	    return self.run(*args, **kwargs)
	  File "/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tasks.py", line 39, in save_voxelwise_pearson_similarity
	    transformation="voxelwise")
	  File "/opt/nv_env/local/lib/python2.7/site-packages/django/db/models/manager.py", line 92, in manager_method
	    return getattr(self.get_queryset(), name)(*args, **kwargs)
	  File "/opt/nv_env/local/lib/python2.7/site-packages/django/db/models/query.py", line 357, in get
	    self.model._meta.object_name)
	Exception: Similarity matching query does not exist.

infocortex:
I'll add it to puppet, but for the moment you have to run
`scripts/add_pearson_similarity_metric.py` manually.  (It generates the
Similarity DB record even if there's no images to compare yet.)

On Mon, Feb 9, 2015 at 7:45 PM, Chris Filo Gorgolewski <

> I just created a fresh virtual machine, created a new user, added a new
> collection and three images in it. Since I could not see the comparison
> button I checked celery with "./manage.py celery worker -A
> neurovault.celery" I see the following error:
>
> [2015-02-10 04:44:31,990: ERROR/MainProcess] Task neurovault.apps.statmaps.tasks.save_voxelwise_pearson_similarity[2619bfaf-6b27-4bbe-96a2-20751eecda31] raised unexpected: DoesNotExist('Similarity matching query does not exist.',)
> Traceback (most recent call last):
>   File "/opt/nv_env/local/lib/python2.7/site-packages/celery/app/trace.py", line 240, in trace_task
>     R = retval = fun(*args, **kwargs)
>   File "/opt/nv_env/local/lib/python2.7/site-packages/celery/app/trace.py", line 438, in __protected_call__
>     return self.run(*args, **kwargs)
>   File "/opt/nv_env/NeuroVault/neurovault/apps/statmaps/tasks.py", line 39, in save_voxelwise_pearson_similarity
>     transformation="voxelwise")
>   File "/opt/nv_env/local/lib/python2.7/site-packages/django/db/models/manager.py", line 92, in manager_method
>     return getattr(self.get_queryset(), name)(*args, **kwargs)
>   File "/opt/nv_env/local/lib/python2.7/site-packages/django/db/models/query.py", line 357, in get
>     self.model._meta.object_name)
> Exception: Similarity matching query does not exist.
>
> —
> <https://github.com/NeuroVault/NeuroVault/pull/111#issuecomment-73639738>.
>



-- 
*Gabriel Rivera*

InfoCortex UG
rivera@infocortex.com

+49 (0) 176 7223 1002
+1 917.375.0045

Friedensstraße 11/Kaiserplatz
D-60311 Frankfurt am Main
www.infocortex.com

PR #119, opened 2015-02-04
hotfix: edit images

infocortex:
Fixed handling of errors on a single form (the formset doesn't send the blank extra form when there is an error.

Also added navigation to any form with an error on page load.


chrisfilo:
thanks!

PR #123, opened 2015-02-08
Adding Cognitive Atlas Task/Contrast Selector to Edit Images

vsoch:
This will add cognitive atlas tagging to NeuroVault. A demo of what it looks like is [available here](http://www.vbmis.com/bmi/share/cogatlas/cogatlas.html). A few notes on the interface:

The dropdown is generated with d3 from json data embedded in the page
The function to generate the page is specific for django crispy forms - the user provides a field, and the field input is replaced with the d3 input, for example:

       django_field = "contrast_definition_cogatlas"
       html_snippet = cogatlas.contrast_selector_django_crispy_form(django_field=django_field,include_bootstrap=False)

The dropdown provides the user with a mouseover link to view a description of the task, and a link to add a term.  I chose this strategy because some of the descriptions are incredibly long, and would not render well in a box on the page, for example.  A scrolling box would be extra, un-needed clicks.

There is always a None / Other option.

When the user selects a contrast, the contrast id is saved into the contrast_definition_cogatlas variable of the image (or whatever django_field has been specified):

adding_contrast

When the user selects None / Other, the field returns to it's default None:

contrast_none

 I have [added a script](https://github.com/vsoch/NeuroVault/blob/enh/cognitive_atlas_query/scripts/generate_cogatlas_json.py) to generate the json that can ultimately be turned into a celery task.
 
Chris - I tried to add arrow key functionality to navigate through the list, but I couldn't pull it off. We can add this as an enhancement, I have added an issue to cognitive atlas so I don't forget. 

Finally, my Google skills failed me to figure out how to install a package from a subfolder of github, so I did not yet add the cogatlas api to the requirements.txt.

#### The cognitive atlas querying works as follows:

The [cognitive atlas API](https://github.com/vsoch/cogat/tree/master/api-python) in python that we are developing queries the cognitive atlas to generate a json with tasks matched to contrasts, and a description to embed in the page. While it would be ideal for this to happen real time in NeuroVault (which the function does by default), getting the different data sources and combining takes about 20 seconds, and so as an option, the function can take in a pre-generated file, as I've [implemented in NeuroVault](https://github.com/vsoch/NeuroVault/blob/enh/cognitive_atlas_query/neurovault/apps/statmaps/views.py#L251).

This has several issues. We would want a user to be able to add a term, and then return to NeuroVault to find it. My thinking is that we have two options: 

 1. the Cognitive Atlas API (or my code) can be sped up (I tested parsing the triples vs rdf, and the first is faster, but suggestions [here](https://github.com/vsoch/cogat/blob/master/api-python/cognitiveatlas/utils.py#L86) would be welcome! This is the first time I'm working with rdf, and yes, there are for loops involved :scream_cat: )
 2. generation of the input json can be done via a celery task, and done at some frequency / with some trigger.

#### A strategy for already existing maps
When we release this, we should announce a lottery to win a NV t-SHIRT (with our new logo!) for all people that tag their data. We can simply pull all the tagged entries on some date, ascribe with a user name, and write a script to select 1+ at random. Each single tag of an image == one entry for the researcher! This may incentivize some early tagging, although we need to make sure the cognitive atlas is ready, and also encourage the user to add tasks / contrasts that are not present.


chrisfilo:
Thanks I'll have a look soon. One think I was thinking about is that (correct me if I'm wrong) Cognitive Atlas terms apply only when subjects performed tasks. For structural scans (for example VBM studies) and resting state they do not make much sense - is that right @poldrack?

vsoch:
Great! For those cases I think that it's reasonable for the user to select "None."

poldrack:
Correct

chrisfilo:
I've merged image comparison so you can turn the ontology parser into celery task. You might also consider moving the cogatlas api to a separate repository. I'm not sure the CogAtlas API is the right place to put the code for generating forms. It's not really portable (it has neurovault related things hardcoded), and it makes it hard to figure out where do I need to look to change the form appearance.

vsoch:
I put it there because my original [tool to annotate](www.vbmis.com/bmi/project/cogatlas/annotate.html) images using the cognitive atlas I thought should be part of the API. If it were a more generalized function to create a form from any API/json (which I suppose it could be with work) then elsewhere might make sense, but it's really just for the cognitive atlas. It doesn't actually have any neurovault things hardcoded (well actually there is one "span11" tag that I needed to change to something general like "body.") It should work with any django crispy form - [if you look](https://github.com/vsoch/cogat/blob/master/api-python/cognitiveatlas/html/cognitive_atlas_contrast_selector.html) the data and the django fields are dynamically replaced with the user's variables.  So if you want to change the form's appearance, you would most likely change the form template page in django, specify a different variable field, or change the template I linked above.  @poldrack, what are your thoughts? Does it hurt to have it there?

chrisfilo:
Ok sorry - I just saw this:
https://github.com/vsoch/cogat/blob/master/api-python/cognitiveatlas/views.py#L124
and got confused.

vsoch:
That's a different function for the annotation tool that is specifically for NeuroVault --> brainspell --> cognitive atlas.  That hard-coding should be removed, although I haven't done it yet because it's not high on the priority list. I'm probably the only one that will use that function any time soon :)

neurovault/apps/statmaps/views.py
```diff
@@ -246,7 +246,12 @@ def edit_image(request, pk):
     else:
         form = form(request.user, instance=image)
 
-    context = {"form": form}
+    # Retrieve cognitive atlas contrast selector autocomplete
+    django_field = "contrast_definition_cogatlas"
```
chrisfilo: This breaks for formsets where the field is "image_set-0-contrast_definition_cogatlas", "image_set-1-contrast_definition_cogatlas" etc.

vsoch:
I don't understand - under what context would a field name be "image_set-0-contrast_definition_cogatlas" ? Don't the field names coincide with the module field names in the database tables?

chrisfilo:
Not for formsets (because there are many images on one page there are many fields). Have your tried your autocompletion function in "Edit images" rather than editing a single image?

vsoch:
Oh, no I had only intended for a single edit, and matter of fact I wasn't even familiar with the idea of a "formset" ! I will need to do a version that works for formsets then!  Thanks for the heads up. :L)

PR #124, opened 2015-02-10
Update restframework

jbwexler:
also includes the api guide page


chrisfilo:
Good job! Swagger looks nice! A few things:

 - "try it" option for atlas queries does not work (because swagger does not allow you to include the input parameters)
 - I could not find the query by DOI for collections

jbwexler:
do u think the "try it" option is important to fix? also which query are you referring to? does it use rest framework?

chrisfilo:
The way it is broken right now is quite ugly. We can just disable it if you
prefer it.

I'm talking about this query:
http://neurovault.org/api/collections/?DOI=10.1016/j.neurobiolaging.2012.11.002

chrisfilo:
did you forget to add neurovault/apps/statmaps/tests/test_data/api/DorsalFrontal_thr25_1mm.nii.gz ?

PR #162, opened 2015-02-22
find_similar page improvements to fix #148, more spacing, two lines of t...

### LINKED ISSUE #148: find_similar  page improvements
* Write the name of the map and collection it belongs to on top
* Dedicate more space for the names of the maps presented

vsoch:
This will do slight tweaks to the "find similar" interface, including:

- Write the name of the map and collection it belongs to on top
- Dedicate more space for the names of the maps presented

The image collection name is displayed at the top of the image, and the name on the bottom, next to the score. The query image is also titled, with the title linking back to the query image. The updates to pybraincompare can be [viewed here](https://github.com/vsoch/pybraincompare/commit/bd9ffd491ef346e72eae918ed43217a6cf0a69cc)

### DIFF
--- a/neurovault/apps/statmaps/templates/statmaps/compare_search.html
+++ b/neurovault/apps/statmaps/templates/statmaps/compare_search.html
@@ -20,6 +20,7 @@ var addthis_config = {
 var portfolio_filter = document.getElementsByClassName("portfolio-filter")
 var li = document.createElement("li");
 var a = document.createElement("a");
+var image_a = document.createElement("a");
 
 function setAttributes(el, attrs) {
   for(var key in attrs) {
@@ -28,8 +29,13 @@ function setAttributes(el, attrs) {
 }
 
 setAttributes(a, {"class": "btn btn-warning", "disabled": "true"});
+setAttributes(image_a,{"style": "float:right; margin-right:40px; font-weight:800; color:LightSeaGreen;", "href": "{{ image_url }}"});
+
 a.appendChild(document.createTextNode(["{{ images_processing }} images still processing"]));
-portfolio_filter[0].appendChild(li.appendChild(a))
+image_a.appendChild(document.createTextNode(["{{ image_title }}"]));
+li.appendChild(a)
+portfolio_filter[0].appendChild(li.appendChild(image_a));
+
 
 $(document).ready(function() {
 
--- a/neurovault/apps/statmaps/templates/statmaps/compare_search.html
+++ b/neurovault/apps/statmaps/templates/statmaps/compare_search.html
@@ -36,16 +36,6 @@ image_a.appendChild(document.createTextNode(["{{ image_title }}"]));
 li.appendChild(a)
 portfolio_filter[0].appendChild(li.appendChild(image_a));
 
-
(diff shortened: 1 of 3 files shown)


neurovault/apps/statmaps/templates/statmaps/compare_search.html
```diff
@@ -28,8 +29,13 @@
...
+setAttributes(image_a,{"style": "float:right; margin-right:40px; font-weight:800; color:LightSeaGreen;", "href": "{{ image_url }}"});
+
 a.appendChild(document.createTextNode(["{{ images_processing }} images still processing"]));
-portfolio_filter[0].appendChild(li.appendChild(a))
+image_a.appendChild(document.createTextNode(["{{ image_title }}"]));
+li.appendChild(a)
+portfolio_filter[0].appendChild(li.appendChild(image_a));
+
 
 $(document).ready(function() {
```
chrisfilo: please remove it not used

vsoch: what do you mean?

chrisfilo: this function seems to be empty

vsoch: Oh i see, that was there from some previous version - not sure why I left it.

chrisfilo:
Looks good, but could you use the same colors (black or blue) and fonts as the rest of the website? Thanks!

vsoch:
https://www.youtube.com/watch?v=dyiIQzzkjVE

chrisfilo:
I apologise in advance for paying attention to minor details so much, but the hover is still orange (instead of an underline) and the fonts still look different. It would be great if you could fix this unless it takes too much time.

vsoch:
Details are important! I forgot to look at the styling that is embedded with the similarity search (from within pybraincompare)! I might switch to standard bootstrap and minimal other styling... need to think about it.

vsoch:
Ah, I know what I should do. I need to separate the "non standard" style elements from the interface and have include/not include it be an input arg.  Then the user can specify not include, and have his/her style determine the page.

vsoch:
ok, styling has been flushed out to [include only essentials](https://github.com/vsoch/pybraincompare/commit/7d90af2f3c81a5c59f3af701fdb57c0a99525882), and eliminated are customizations that are "obvious" (font style, weight, color, and link behavior). The page should default to the user's style first, and then bootstrap. This is what it looks like for my test:

cookies

I didn't think that font weight had such a salient role in the overall look, but it really does. The one font weight that I kept at a slightly modified (larger) value was the title for the query image. That font weight is actually in the [neurovault code base](https://github.com/vsoch/NeuroVault/blob/enh/image_comparison_adv/neurovault/apps/statmaps/templates/statmaps/compare_search.html#L32) and not pybraincompare, and could be modified easily.

My glass brains are all the same (dummy images I created) but I am still painfully distracted now by the fact that the color scale is a maximum intensity projection that does not properly represent negative values, or even value scales between images! Ohma.

PR #197, opened 2015-04-15
Update/image comparison Tweaks

### LINKED ISSUE #133: Image comparison needs more tests
The results I see seem unlikely (when comparing the r score to the scatter plot). This might be due to other bugs, but none the less we need tests on simulated data (when we know what correlation to expect) for the comparison methods.

vsoch:
This pull request should do some tweaks to the initial update/image comparison:

- The image counter was not accurately counting the images still processing. We now have functions in utils to both count and get images processing and existing, taking into account StatisticMaps and NIDMResultStatisticMaps
- There was a format character error within pybraincompare that was just fixed (this error would only generate in Opbeat and we should check if/when the production is updated)
- There is now a warning on scatterplot atlas page if either of the images is thresholded.

We also should close #133 - work done from previous PR.

### DIFF
--- a/neurovault/apps/statmaps/templates/statmaps/compare_images.html
+++ b/neurovault/apps/statmaps/templates/statmaps/compare_images.html
@@ -10,6 +10,12 @@ var addthis_config = {
 {% endblock %}
 {% block content %}
 
+    {% if warnings %}
+       {% for warning in warnings %}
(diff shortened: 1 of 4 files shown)


vsoch:
For pybraincompare - I'll actually need to push the new version to pip (not done yet)

vsoch:
ok, now is updated on pip.

vsoch:
:green_heart:

PR #218, opened 2015-04-24
Fix API by adding pagination

### LINKED ISSUE #122: paginate /api/images
There are now enough images that requesting /api/images takes a while to return data (~1 minute). Can we paginate this so that only the first 100 results are returned by default, and then the user has to pass a paging key for subsequent images?

vsoch:
This is a quick fix to add pagination to the API (and fix it's broken-ness!). It was incredibly simple to implement and get working. A few notes:

- The default urls (eg "neurovault.org/api/images") will return 100 results per page, starting at offset 0.
- The user can set the ?limit= argument to return more results, and we set a max_result limit of 1000
- To accomplish this, we have a custom class defined in the [urls.py](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/apps/statmaps/urls.py#L147), and we set this to be the default behavior in [settings.py](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/settings.py#L219)
- this update is intended to fix #122 and does not improve performance of the API. Reading [here](http://www.dabapps.com/blog/api-performance-profiling-django-rest-framework/) we most definitely want to use cache - I will open a new issue for this.

### DIFF
--- a/neurovault/apps/statmaps/models.py
+++ b/neurovault/apps/statmaps/models.py
@@ -262,8 +262,7 @@ class Image(PolymorphicModel, BaseCollectionItem):
         try:
             url =  self.thumbnail.url
         except ValueError:
-            url = os.path.abspath(os.path.join(neurovault.settings.BASE_DIR,
-                                 "static","images","glass_brain_empty.jpg"))
+            url = os.path.abspath(os.path.join("/static","images","glass_brain_empty.jpg"))
         return url         
 
     @classmethod
--- a/neurovault/apps/statmaps/models.py
+++ b/neurovault/apps/statmaps/models.py
@@ -19,6 +19,7 @@ from gzip import GzipFile
 from django import forms
 from xml import etree
 import nibabel as nb
+import neurovault
 import urllib2
 import shutil
 import os
@@ -256,6 +257,14 @@ class Image(PolymorphicModel, BaseCollectionItem):
             return_args.insert(0,str(self.collection.private_token))
             url_name = 'private_image_details'
         return reverse(url_name, args=return_args)
+
+    def get_thumbnail_url(self):
+        try:
+            url =  self.thumbnail.url
+        except ValueError:
+            url = os.path.abspath(os.path.join(neurovault.settings.BASE_DIR,
+                                 "static","images","glass_brain_empty.jpg"))
+        return url         
 
     @classmethod
     def create(cls, my_file, my_file_name, my_name, my_desc, my_collection_pk, my_map_type):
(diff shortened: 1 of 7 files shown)


vsoch:
As soon as this is live I will update the python api.

vsoch:
Also, I couldn't figure out the styling for the API components - my local version looks different than what is live:

nostyle

Depending on where the production styling is coming from, we may need to style the pagination buttons. If we could talk about how the styling is done, it would be very helpful, and I'll do the styling.  :sandal:

vsoch:
was rest_framework removed from static? I might have an old version that stuck around.

https://github.com/NeuroVault/NeuroVault/tree/master/neurovault/apps/statmaps/static

vsoch:
ah, figured it out, I just needed to collectstatic :)  Looks good!

looksgood

vsoch:
and bootstrap-tweaks is not there because it is copied from the rest_framework package folder:

       Copying '/opt/nv_env/local/lib/python2.7/site-packages/rest_framework/static/rest_framework/css/bootstrap-tweaks.css'

chrisfilo:
Works as advertized. I have two problematic issues:
1) if you don't specify "limit" (hitting http://192.168.33.10/api/collections/?format=json) you get all collections/images - this will cause server timeouts. To avoid this we should make pagination compulsory.

2) It seems that max_limit parameter is ignored 
screen shot 2015-04-25 at 11 53 36

vsoch:
Hmm, it could be because page_size needs to be PAGE_SIZE, let me test this.

vsoch:
Quick question - for the result that shows 1019, are you sure that you aren't being shown 1000? For me, the count will show the total in the query, but the pagination is working (returning only N on the first page, and total-N on the second).

vsoch:
and it looks like we should set PAGE_SIZE and default_limit, let me push those.

vsoch:
ah, the api_tests are failing now! At least that means the default_limit on the pagination is likely working :)

vsoch:
I need to do morning stuffs but I can look at this in a couple hours, unless you look first!

vsoch:
Ok, this should fix the test. The change is that now when we do a query with a url asking for a specific model, eg:

     url = "/api/images"

there is one more level to the json before the results:

      In [154]: response.keys()
      Out[154]: [u'count', u'previous', u'results', u'next']

I've also added a function within image to return the thumbnail url, if it is defined, and if not, to return the dummy brainmap image url, so we don't get the file field not defined error.

PR #223, opened 2015-04-25
API: Remove "images" and "nidm_results" from being returned with collection

### LINKED ISSUE #222: Collections API should not return all image in a collection
The API currently returns a list of images associated with a collection, and this is not a reasonable thing to do as the collection with 6000 images will time out for a single query, eg:

      http://neurovault.org/api/collections/?format=json&limit=1&offset=157

I put the link in a code box because you should not click it :)

The user should need to retrieve the collection first, and then get the images with the image api (paginated) based on the collection.

vsoch:
This will fix #222


chrisfilo:
After removing these endpoints is very hard to obtain all images from one collection without downloading the metadata of all images. We should add #224 if we want to do this.

vsoch:
I've added the following to this PR:

### Joblib to load pickle
We now load pickle files using joblib from sklearn to slightly improve speed. This is done when [calculating comparisons](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/apps/statmaps/tasks.py#L142), and loading pickles for the [compare_images view](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/apps/statmaps/views.py#L741), and ]one other instance](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/apps/statmaps/views.py#L672).

### API addition for collection images
The API has an added path /api/collections/PK/images to return images. I made this work by [dynamically removing the "images" and "nidm_maps,"](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/urls.py#L342) and then adding back in.

### Convert Zeros to NaNs before transformation
To generate the transformations, we first set all 0 values to NaN. Since this is only possible with data that is float, I create an empty numpy array (data_type float64) first to put the data. If I don't do this, if there is an image with int16 type, there will be an error. Then since we resample with default interpolation = "continuous" this means that the value vectors will NOT have absolute 0s resampled to "almost zero," and the NaNs will be maintained. Then when we calculate the pairwise deletion mask, the NaN values will not be included. This has substantial impact (already I can see) on the similarity scores - since we don't include these zero values they will likely go up.  However, this fix should resolve the issue of the vertical and horizontal lines in the scatterplot compare, and we would need to recalculate all previously calculated transformations. Please [take a look at this section](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/apps/statmaps/tasks.py#L68) of the code and check that my strategy is logical.

### Updated test_comparison tests
I [added a test to test_comparisons](https://github.com/vsoch/NeuroVault/blob/fix/api/neurovault/apps/statmaps/tests/test_comparison.py#L79) to make sure the NaN values in an image are maintained when the transformation is calculated. Since we did not have test data with substantial NaNs, I added a version of the motor_lips.nii.gz to the test data with substantial nan values. This will be useful for future tests as well. I also updated the test_comparison reading images (first section) to use functions from utils.

chrisfilo:
Great!

The way you implemented it /api/collections/PK/images returns collection metadata and list of images - the same way the old version used to (see: http://neurovault.org/api/collections/504). I see two problems with it:
1. It's a bit counterintuitive - as a user I would expect it to list just the images without the metadata. Just like a slice of /api/images but for one collection.
2. It does not take advantage of paging so big collections will time out. See: http://neurovault.org/api/collections/503

Also please update docs with new API paths.

vsoch:
We have docs?

chrisfilo:
http://neurovault.org/api-docs

vsoch:
oh man...

vsoch:
I can't figure out how to serialize the nested images

chrisfilo:
what about something along the lines of:
        
	@detail_route()
	def images(self, request, pk=None):
		collection = get_collection(pk,request,mode='api')
		queryset = collection.image_set
		serializer = ImageSerializer(queryset, many=True)
		return Response(serializer.data)
		
		
Just a rough sketch - I have not tested it.

vsoch:
Ok, api-docs are updated, and only images are returned. I couldn't figure out how to make the pagination happen for the images... sorry :disappointed:

chrisfilo:
Check out the "recent_users" example here: http://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing

neurovault/apps/statmaps/views.py
```diff
@@ -670,7 +669,7 @@ def atlas_query_region(request):
         else:
             synonymsDict = {}
             with open(os.path.join(neurovault_root, 'neurovault/apps/statmaps/NIFgraph.pkl'),'rb') as input:
```
chrisfilo: Does joblib.load work with this file? It was not created with joblib.

vsoch: I think so, I would have gotten a lot of errors if that weren't the case, and here is a quick test:

      import pickle
      from sklearn.externals import joblib
      data = [1,2,3]
      pickle.dump(data,open("/tmp/pickle.pkl","wb"))
      joblib.load("/tmp/pickle.pkl")
      [1, 2, 3]

Issue #225, opened 2015-04-27
Last Two Collection Links Under API-Docs Page are Missing an "s"


vsoch:
This is true for the text for the link and the address itself. For example, this:

      http://neurovault.org/api/collection/287

should be this:

      http://neurovault.org/api/collections/287

and all other links checked as well!

PR #226, opened 2015-04-28
Refactoring of reduced representation

chrisfilo:
* Switch to numpy from joblib (same performance, but allows using File objects)
* Use FileField instead of paths

### DIFF
--- a/neurovault/apps/statmaps/migrations/0045_auto_20150428_0229.py
+++ b/neurovault/apps/statmaps/migrations/0045_auto_20150428_0229.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import neurovault.apps.statmaps.models
+import neurovault.apps.statmaps.storage
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('statmaps', '0044_auto_20150423_2134'),
+    ]
+
+    operations = [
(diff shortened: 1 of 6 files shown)


chrisfilo:
@vsoch could you have a look?

vsoch:
Sure, I'll test it locally as well... give me a few mins!

chrisfilo:
Only two underscores in a row are problematic: https://docs.djangoproject.com/en/1.8/topics/db/models/#field-name-restrictions

vsoch:
Ok, it looks like single is just for [hidden field "id"](https://stackoverflow.com/questions/14945242/django-query-single-underscore-behaving-like-double-underscore)

vsoch:
What is rationale for moving the reduced_representation back to Image? We had moved it to BaseStatisticalMap for #205 

Other than that, everything seems ok!

 - celery tasks work for new images and updates
 - adding collection with new images worked
 - scatterplot compare and find similar views don't have error
 - api is good!

vsoch:
We maybe should fix #225 on this PR since people's attention was called to the updated API. If not I can add it to my next batch of work.

chrisfilo:
I think we will be able to use reduced_representation to calculate similarity between different types of images in the future and that's why I moved it there. The actual similarity calculation as it stands right now is applicable only to statistical maps so it should stay there, but the reduced_representation file can have more broad uses in the future.

vsoch:
ah, good thinking.

PR #184, opened 2015-04-07
WIP: Metadata support for images

burnash:
This is an implementation of arbitrary key-value fields for images #163.

This pull request is more of a code review than a real PR, as some implementation concepts are subject to change.

To test it locally:
 
 1. Install hstore extension on the database:  `create extension hstore;`
 2. Run `./manage.py migrate`

The key-value storage of choice is hstore as it indeed has more mature support from Django.

I've put the hstore field into the `Image` model, but I'm not sure if it's a best place for it. All data handling happens in `neurovault/apps/statmaps/views.py`. It might be a good idea to move it to DRF. There are two additional endpoints in `urls.py`:
 *  `collection/<cid>/importmetadata`
 *  `collection/<cid>/editmetadata`

Basic backend tests are in `neurovault/apps/statmaps/tests/test_collections.py`

Front-end uses [Handsontable](https://github.com/handsontable/handsontable) to display a data grid. [PapaParse](https://github.com/mholt/PapaParse) to parse input CSV files and [Fuse](https://github.com/krisk/fuse) to assist in fields mapping.

 * `dataimport.js` extends Handsontable to allow a user to map own fields to some predefined fields. It also provides miscellaneous validation procedures.
 * `metadata-import.js` handles file open, CSV parsing and definition of fields and DataImport init.
 * `metadata-edit.js` displays a data grid with image metadata.

Issue #171, opened 2015-03-10
Option to apply same settings to all images in collection


chrisfilo:
From: https://neurovault.uservoice.com/forums/210717-general/suggestions/7189722-option-to-apply-same-settings-to-all-images-in-col (and also @GaelVaroquaux )

When having many images in a collection (e.g. uploaded from folder) it would be easier to have an option to set all images to same Map type/ Modality / Cognitive Paradigm.

tyarkoni:
Per discussion in #234, one approach that would solve this problem would be to have a separate ImageMetadata class that can be attached to both Images and Collections, with the latter case serving as the default for all images. Then there could be an "apply to all" button at the Collection level that overwrites individual image metadata records.

Issue #163, opened 2015-02-25
Support for extra key-value fields


chrisfilo:
It would be useful for some purposes to add an ability to annotate your maps with fields that are not in the fixed, predefined set. This is especially relevant in context of single subject maps and population/behavioural variable related to them.

From the user perspective this feature should consist of the following components:
1. Ability to upload a folder or .zip archive that consists of a series of maps and a CSV file with the following structure:

	filename, field_1, field_2, field_3
	map001.nii.gz, 23, 56
	map002.nii.gz, 54, 78
	map003.nii.gz, 77, 87
	
2. Ability to view and edit the new key-value fields.
3. Ability to add new key-value fields from the user interface.

In terms of implementation PostgreSQL provides two options at the moment: hstor and jsonb. Hstor is slower, but more mature and has good django support (https://github.com/djangonauts/django-hstore) including REST API extensions (https://github.com/djangonauts/django-rest-framework-hstore). JSONB on the other hand is faster (even faster than MongoDB!), but is relatively new. It has only been added to PostgreSQL 9.4 which was only released in December 2014. Some preliminary django support exists, but is not mature: https://pypi.python.org/pypi/django-pgjsonb/0.0.10

chrisfilo:
In addition after uploading the data user should see a website on which they can map their variables to set of predefined commonly used terms (such as age, handedness etc.) This way we will maintain flexibility as well as minimal correspondence between the terms.

PR #234, opened 2015-05-04
WIP: Support "fixed" image fields in metadata import

burnash:
This is my take to address UX consistency issues described in @chrisfilo's [comment](https://github.com/NeuroVault/NeuroVault/pull/184#issuecomment-94373548) to #184 :

* "Fixed" metadata fields in statistical map objects are updated through the CSV upload in the same way as custom metadata fields.
* In the API, custom metadata fields are exposed in the same way as "fixed" fields.

My initial approach was to replace "fixed" fields in models with "virtual" fields described in Images model using hstore's [schema mode](https://github.com/NeuroVault/NeuroVault/pull/184#issuecomment-94580050) or some custom "virtual" fields implementation. This led to numerous inheritance errors and conflicts with PolymorphicModel. 

In this branch I took a different, more explicit and more flexible way to solve the problem.

1. Each model that needs to define "fixed" fields implements a class method `get_fixed_fields` which returns a tuple of field names.
2. Upon saving metadata key/value fields in database the code checks if a key corresponds to a "fixed" field and saves this field to a model if it does. This works both for import a new CSV file and for editing metadata as a grid.
3. When serializing an image model for API response, the code merges "fixed" metadata fields with custom metadata fields giving precedence to the former.

This PR is branched from #184 all modifications are in d71705b.


chrisfilo:
Thanks for the update. Looks great! I listed some comments below

1. Please update the puppet config to include hstore (you will need to install postgresql-contrib)
3. The UI seems to be cluttered at the current stage. I would propose combining Edit Images and Edit metadata into one dropdown button http://getbootstrap.com/components/#btn-dropdowns @vsoch @tyarkoni might chip in how to improve the interface.
4. There is no validation for the values added in the table form. For example image modality takes only a few predefined strings. If user inputs something else in the tabular interface the page will not complain, but API calls break (KeyError). Please add a test for this scenario.
5. cognitive_paradigm field is missing in the tabular interface.
6. It might be useful to add an example file in the csv upload instructions
7. If possible the first column of the tabular interface (File name) should be locked for edits.
8. "description" and "subject_id" fields should not be compulsory.
9. "Details" and "metadata" tabs should be merged
10. Having both "image type" (group or single) and "map type" (T, Z, etc.) is confusing. I agree we need to clearly indicate that a map is a single subject map. What about changing "image type" to "number of contributing subjects" (numerical value). This way we could kill two birds with one stone - distinguish group level from single subject and get sample size. @ljchang what do you think?


PS For people updating their VMs - you need to:
	1. sudo apt-get install postgresql-contrib
	2. sudo su - postgres
	3. psql neurovault
	4. CREATE EXTENSION hstore;

vsoch:
Good work! And thanks for the VM commands @chrisfilo.  The easiest solution is to do a [button drop down](http://getbootstrap.com/components/#btn-dropdowns), which I see that Chris already linked. I just searched around looking at other options, and none are as straightforward/simple as this.  Any kind of re-haul of the navigation should only be done when/if we redesign the entire UI, which is not happening now.

It would even clean up the interface quite a bit to have one "edit" button with dropdown to images, collection, and metadata.  The same might be done for upload - it's a lot easier for the user to have one choice for upload, and then clearly see the two upload options (versus scanning all the buttons to see if he/she missed some other upload option.) I don't think it should be an issue that we can't navigate from within either collection or edit images to the other, because editing either is a pretty rare event.

ljchang:
Nice work Anton!

I agree that it would be nice to simplify the UI a bit with a dropdown for editing and uploading.

I also agree that having "image_type" and "map_type" is confusing.  We would also like to add an additional type for "pattern_maps".  This will allow us to test pattern responses on various datasets on the new analysis website, similar to the neurosynth decoding.  I could also imagine having users upload "meta_analysis_maps" at some point too.  Not sure if that helps justifies the additional field.  It's not a bad idea to use the sample size instead of group_image, but this might be a bit confusing for users to remember especially if they are only uploading group maps as the header will still be something like "Image_type".

vsoch:
We already have a substantial number of meta analysis maps from NeuroSynth, GingerAle, etc, for example:

- [Collection 7](http://neurovault.org/collections/7/)
- [Collection 13](http://neurovault.org/collections/13/)
- [Collection 73](http://neurovault.org/collections/73/)
- [Collection 439](http://neurovault.org/collections/439/)
- [Collection 474](http://neurovault.org/collections/474/)

Having a field to denote the number of subjects to replace map_type would do well to distinguish individual vs. group maps, but it wouldn't solve the problem of distinguishing these meta analyses maps from others - a distinction I think is important to make for 1) more advanced future analyses, and 2) for the user to filter these maps from the set to do a meta-analysis.  The "map_type" is a separate thing entirely (eg, a meta analysis can be a Z score map) and I agree "image_type" is too similar, and it's confusing. What about "analysis_type" or something along those lines?

chrisfilo:
Ok I think the most sensible would be the following:

* replace image_type (not map_type) with number_of_subjects (a numerical field); remove number_of_subjects field from Collection
* add a from_meta_analysis boolean field

Would that work for everyone? @tyarkoni ?

PS @ljchang I did not understand what you meant by saying 'as the header will still be something like "Image_type"'

vsoch:
Is the desire for the Boolean to avoid checking strings? If we choose this strategy, then if another "analysis_type" becomes important to distinguish, it would mean adding another Boolean. Does that seem reasonable? For example, we have a lot of parcellations, some VBM, lots of connectivity, etc. At some point it would be nice to be able to filter those.

ljchang:
@vsoch beat me too it, haha.  I was going to say the same thing.

ljchang:
Also, @chrisfilo would did you mean by point 4 (cognitive_paradigm) field?  Do you mean the task that was used?  Cognitive process probed?

ljchang:
I think I like the "analysis_type" strategy that could encompass meta-analysis, single-subject, univariate group analyses, pattern maps, parcellations, etc.  This would make it easy to filter for many different types of analyses without requiring multiple boolean fields.   However, we should probably only have a finite number that have predetermined labels (that can be checked) to ensure consistency.

vsoch:
Yes! That is [very possible](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L423). :0)

tyarkoni:
I like the idea of making number_of_subjects editable at the image level (if that's the idea). Not sure about removing it from Collection though; it seems like most collections are currently multiple contrasts from the same study, and it would be kind of a pain in the ass to have to specify this repeatedly for every image.

In general, I think a good way to handle this kind of situation (and I apologize if this has been discussed before, I haven't been following closely) would be to have an ImageMetadata model that can be joined to either an Image or a Collection, where the collection-linked record specifies defaults for all included images. That way, a user could edit the same metadata at either level (probably using the same templates), and they would only need to edit at the Image level for any fields that differ from the Collection default. Looking at the current Collection metadata, I would say probably 90% of it could be handled this way, as it is mostly stuff that theoretically *could* vary on an image-by-image basis within a single collection.

tyarkoni:
As far as the image_type thing goes, I think handling it like the modality field @vsoch linked to makes sense. At some point it might make sense to create a separate model for ImageType (e.g., if you plan to start supporting similarity operations on some image types but not others, or linking meta-analysis maps to specific studies in Neurosynth, which would be easily doable)--but that's probably overkill at the moment.

chrisfilo:
I'm not sure about "analysis_type":

*I admit there is a need to identify parcellations (which can be done automatically) and maps coming from meta analyses (which requires user input).
* analysis_type seems like a broad term that will have different options depending on the modality ("VBM" does not make sense for fMRI for example; someone might want to have a parcellation of a single subject)
* it cannot be a free form field - we need options: should we list all analysis techniques in neuroimaging?
* what is the difference between meta analysis map and a pattern map?
* if it's a group map or meta analysis map the number of subjects is one of the most useful piece of meta data - we should still try to get it

@tyarkoni we should revisit metadat common across images here: https://github.com/NeuroVault/NeuroVault/issues/171

@ljchang: cognitive_paradigm: https://github.com/NeuroVault/NeuroVault/blob/17b7c089e6afaabe59a9b29e0c228a6e45070a67/neurovault/apps/statmaps/models.py#L444

nicholst:
Hi folks,

Just to amplify that, if you haven't already noticed, the ALE maps are strange beasts.  The units are totally different from Z or anything (see e.g. [Collection 474](http://neurovault.org/collections/474/)'s uploaded ["cognitive control ALE"](http://neurovault.org/media/images/474/updated_cognitive_control_32515_Fixed_k115_FixedFlippedCoords_NoClustThresh_ALE.nii.gz), with non-zero Q1 | M | Q3 of 0.000137 | 0.001821 | 0.006207), and depend on the number of studies combined.  Also note that ALE units depend on the kernel size and voxel grid size used to define the map (though since I guess most people use the software and its defaults, I guess most ALE will at least have consistent units).

I was going to say that MKDA maps are also a different beast (but at least a well-defined binomial proportion beast), but it seems from Tor's [Collection 7](http://neurovault.org/collections/7/)'s [Executive_revInf_z](http://neurovault.org/images/10/), they are transformed to Z.

So... do you want to add a [BaseStatisticMap](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L338) type of 'ALE'?

tyarkoni:
I don't think an analysis_type field would have to encode detailed information about the analysis... it could just be a very general descriptor. It's probably not realistic to shoot for a single, informative machine-readable description that tells us whether, e.g., a meta-analysis was done with ALE, Neurosynth, MKDA, etc. That could get unwieldy very quickly and would require continual updating of available options as people introduce new techniques. But even without detailed information, it would be very useful to know that an image was a meta-analysis of some kind, as opposed to an experimental contrast (and a drop-down selection of this kind seems like the minimum effort that we can expect many/most users to be willing to invest). Obviously some decisions would have to be made about which types of analyses to include as valid options, but as long as there was an "Other" option, I don't think it could hurt to have a field like this.

vsoch:
+1

chrisfilo:
What other options other than ["metanalysis", "other"] should we consider for this field?

vsoch:
I'll give a first go... please add / subtract from this!

["meta-analysis","parcellation","single-subject","univariate-group", "other"]

And I'm not clear about what @ljchang means by "pattern-map,"  but maybe that one too.

Would we allow for more than one tag? For example, a single-subject vs. group parcellation.

ljchang:
This sounds like a good plan.  "pattern-map" refers to multivariate weight maps like Tor's pain signature.  It is the main output of the new machine learning analysis application we are building.  It will be nice to have this so we can filter patterns to apply to new datasets.

tyarkoni:
At the risk of complicating the metadata model, it might be worth having separate fields for "level" and "analysis type". Valid options for level would be, e.g.,[single subject, group analysis, meta-analysis, other]. Analysis type options could be something like [clustering/parcellation, functional connectivity, multivariate analysis, univariate contrast, behavioral correlation, ROI, other]. Some combinations wouldn't make a whole lot of sense, but that's okay--I don't know that we can make these fields mandatory anyway. The defaults could be set to level='group' and analysis_type='univariate contrast', which would probably cover a majority of images.

nicholst:
So what are the units of a 'pattern-map'?  Are the z-ish, or just arbitrary weights?

Just try harping on this because knowing that an image is Z/T is incredibly powerful, and if [BaseStatisticMap](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L338)?  is set wrongly it could screw up any methods that assume use that information (assuming BaseStatisticMap is set correctly).   In general we should warn/suggest to users somehow that 'pattern-map' or 'ALE' images should carefully set Stat type, if only to 'other'.

ljchang:
The weights in the ‘pattern-map’ will be fairly arbitrary and are usually quite small.   It sounds like we should probably just set the stat type to ‘other’ as a default for these images to avoid any issues. Also, this isn’t a big deal but they also don’t really need to be thresholded in the viewer.

ljchang:
Just wanted to summarize the plan so that @burnash can get started on implementing the suggested changes.  I was initially going to argue against @tyarkoni's more complicated suggestion of separating the meta-data into level of analysis and analysis type and vote in favor of @voch's single "analysis-type" variable, but I think I changed my mind when I was writing the summary.  Of course this will be more complicated for users and would also potentially complicate filtering images, but it gives us a lot more flexibility for running analyses on the data, which is the whole point of the image metadata anyway.

1) I propose that we start with something like this:

"analysis-level" : ["single-subject", "group-analysis", "meta-analysis", "other"]
"analysis-type" : ["parcellation", "functional-connectivity", "pattern-map", "univariate-contrast", "behavioral-correlation", "roi", "other"]

2) Do we want to add the "statistic-type" from the collection level meta too? e.g., ["beta","z","t","other"].  Though even more complicated, I think this would fully meet the needs of everyone's suggestions.

3) Related to point 2, we never really discussed@tyarkoni's other suggestion of accessing the collection metadata through the tabular image metadata interface in order to make manual changes.  This would be really useful for inputting and revising collection metadata.  However, there are so many collection metadata fields that this would also make the table overly expansive and many of the fields would probably never be used anyway.  Another consideration is to figure out how useful this would be after an upload API has been implemented.  We (really @burnash ) were planning on helping with this implementation as it is necessary for our analysis-website to populate neurovault with the new pattern maps that result from the machine-learning analyses.  Unless someone feels strongly one way or the other, we could also postpone this discussion until we see how it goes interacting with the tabular data and also potentially the upload API.

4) Finally, I was under the impression that we were going to require the analysis fields in the image metadata, especially since we are including an option for "other".  Does anyone forsee a problem with that? or have a strong opinion?

chrisfilo:
Check out current field "map_type" - it describes the statistic field
already. More comments to come.

vsoch:
I just was going to say the same thing :) [here it is](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L345)

burnash:
Thanks to everyone for the feedback.

Presently I'm iterating through @chrisfilo's comments. I've pushed commits for points 1, 2 & 8.

For the point 2 I'd like to ask @vsoch and @chrisfilo to review my implementation of dropdown buttons on the collection details page. 

Kudos to @vsoch for a detailed comment on this. I've combined buttons for both Edit* and Upload* actions and placed a new Import Metadata button to the Upload group.

Here's how the actions panel looks like:

edit-dropdown

And the Upload:

upload-dropdown

vsoch:
The menu looks beautiful - and it's so much easier for the user to find what he/she is looking for with this simplified version. The only thing I wanted to try was seeing how the buttons looked as the same width as the content, for example, with upload:

menu

(my blues look a little different, possibly due to bootstrap 3) 

I like your version much better!

burnash:
@vsoch Great, thank you!

When will a bootstrap 3 migration happen?

vsoch:
Likely in my next PR (which I imagine will be sometime this weekend or early next week?) and then when that is merged.  But I will wait for this PR to go through first, because I am pretty excited about your new additions! :gift:

chrisfilo:
Thanks for the updates:
* I would keep boostrap 3 update independent and not make PRs wait for each other
* In Edit button I would rename "Collection" to "Collection metadata" and "Metadata" to "Images metadata"
* In upload I would rename "Import metadata" to "Images metadata (.csv)"

I still find it problematic that when someone goes to "Edit images" or "Edit image" views they will not be able to see or edit custom fields stored in hstore.

chrisfilo:
@ljchang I would leave @tyarkoni's suggestion for a separate PR for now.

I'm happy with the "analysis_level" field, "statistic_type" is already covered by "map_type" (I would prefer to keep it this way for backward compatibility. I'm still not convinced about the "analysis-type". Even the examples you gave are not very clear. For example:

* "pattern-map" - I would argue that this is not a very well established term and it is not well defined
* "univariate-contrast" - is seed based analysis of old vs. young a univariate-contrast? roi? functional connectivity?
* "behavioral-correlation" - is that everything else then individual differences?

I think we should really ask ourselves the question what do we need this field for? What kind of queries? What kind of applications?

ljchang:
@chrisfilo agreed, let's leave @tyarkoni's suggestion for a separate PR.

"analysis-level" and "map_type" sound good to me.  The "analysis_type" field is going to be very important for the analysis website.  Though I agree that the semantics could probably be improved.  For example, "behavioral-correlation" could refer to individual difference analysis when paired with "group-analysis", but could also be a parametric analysis when paired with the "single-subject" level.   I totally agree that your example could be a "univariate-contrast" or "functional-connectivity".  I don't think we are ever going to come up with a good ontology of general analysis types as it changes very rapidly.

Here are a few of our specific use cases for the neuro-learn website.

1) **Train multivariate pattern map from neurovault collection**.  This will require users to select datasets available from neurovault and decide if they want to create "single-subject" level "pattern-maps" or "group" level "pattern-maps".  They will then select a metadata variable to use as the training labels.  They will also be able to select the type of algorithm and cross-validation type and also potentially a mask.  The output of the analysis will be a "pattern-map" that predicts or classifies a user defined metadata variable.  The user will have the option of downloading the map or uploading it to neurovault as a new collection.  The analysis provenance information will be included as new metadata. 

2) **Test multivariate pattern map from neurovault collection**.  This will require users to select a "pattern-map" available on neurovault to test on other neurovault collections.  Users will select which datasets they would like to apply the "pattern-map" to and will select the type of test (e.g., dot-product, correlation, cosine similarity, ROC, etc.).  There may be an optional masking step.  The output will be plotted and the results can be downloaded.

3) **Train multivariate pattern map from multiple neurovault collections.**  This is very similar to use case 1, but will involve aggregating data over multiple collections.  We will probably have a method to define new metadata via a tabular interface on the analysis website, not neurovault to avoid bloat.  The rest of the details will be similar to use case 1.

There are many other use cases for this "analysis_type" variable as well.

1) **Find all parcellation images (or ROI)** to use for an analysis or to filter out of meta-analyses.

2) **Don't employ thresholding** when plotting parcellation/ROI/pattern-map images on papaya viewer.  Or use a different colormap (e.g., rainbow) when plotting parcellation/ROI images.  This would be very helpful as these types of images are not being plotted well currently.

I'm sure I could come up with many more examples, but these use cases already highlight a need for metadata that can reveal if the image is ["pattern-map","ROI","parcellation","univariate"].  I'm sure "functional-connectivity", would also be helpful.  I realize that "pattern-map" is not well defined or generally agreed upon and something like "multivariate" could also work, but I don't want unsupervised multivariate analyses ending up with this tag.  Those would be better suited for "parcellation" or "functional-connectivity".

chrisfilo:
Followup question: For the "Test multivariate pattern map from neurovault collection" use case would neuro-learn be only able to work with maps generated by you? Would a group level Z map done with FSL also work?

ljchang:
Hmm... Good question. Only maps trained to predict an outcome would work for some of the tests (e.g dot product, or ROC). However, for a spatial correlation, theoretically any type of map could work as a test map. This would be the equivalent of decoding in neurosynth with other maps, or your new similarity analysis.

chrisfilo:
And what kind of values "pattern-maps" have? probability values [0-1]? positive and negative weights?

ljchang:
positive and negative weights.  We are mostly using linear methods, so they are just beta maps.  The values are typically very small because when you multiply the map by a new data map (3D image) and sum all of the voxels,  you will get a scalar value that is on the same scale as what the weight map was originally trained to predict or classify.  It could be a probability value/log odds (logistic regression), distance from the hyperplane (SVM), or a predicted value (regression).

chrisfilo:
Thanks @ljchang - I now better understand your use cases. Most importantly modification of the fields is independent of the hstore work @burnash is working on. I started a new issue (#263) for it to keep these two things separate. Hopefully it can be addressed in a separate PR. I'm happy to send one if you agree on my proposed solution.

ljchang:
Thanks @chrisfilo, I think your proposal will work well provided "map_type" can be accessed via the tabular image metadata interface.  Though it is technically correct, one minor concern about combining "pattern" with "beta" is that a user might get overwhelmed with "beta" images after filtering.  For example, our giant collection of single trial images are all "beta" images (http://neurovault.org/collections/503).  I'm not sure adding "analysis_level" as an additional filter would help much, because users might upload pattern maps from "single-subjects".

chrisfilo:
I'm happy to make them separate, but we need a clear description of what is
what.

burnash:
@chrisfilo I agree with you.

For me it doesn't feel right as well to have custom metadata separated from
image (or images) metadata in the "edit" mode.

I don't know all the possible use case scenarios for custom metadata. The
examples of image metadata I've seen so far have a large number of fields.
From this I can assume that the most useful representation for this type of
data is a grid UI.

Sure we can add editable custom fields to these views. But I think we need
to make some visual distinction between custom and "fixed" fields. There's
no "verbose names" for custom data, nor "help text". And since these fields
are custom the validation should be different.

chrisfilo:
Sorry for late reply. I totally agree there should be some visual distinction between custom and fixed fields!

vsoch:
I think an important question to ask is, is it important to maintain the distinction if a user were to download (or query programatically) the data? If not, then we can use something simple like color to distinguish. If so, then we need to append something to the variable names (and perhaps still use color). And then there is the option to keep them separate, possibly both in a grid UI but custom tabs. Is there anything implemented that we can look at to think about this?

chrisfilo:
On Sun, May 24, 2015 at 7:54 PM, vsoch <notifications@github.com> wrote:

> I think an important question to ask is, is it important to maintain the
> distinction if a user were to download (or query programatically) the data?
> If not, then we can use something simple like color to distinguish. If so,
> then we need to append something to the variable names (and perhaps still
> use color).
>
I don't think we need to keep them separate in the API.

> And then there is the option to keep them separate, possibly both in a
> grid UI but custom tabs. Is there anything implemented that we can look at
> to think about this?
>
Hmm I can't quite see the solution you are proposing. Could you sketch it
out quickly?

The simplest solution I can think of would be to dynamically generate a
Form based on the content of the hstore.

vsoch:
I don't have a good sense for what @ljchang is proposing, visually, because it seems that you could have hundreds of metadata custom fields, and there is no way that is going to work on the "edit images." page. Instead, I think metadata might have its own tab in the collection view, and browsing to it would show a matrix of the metadata, likely needing scrolling. We would not include this metadata in the edit images form, because that could get ugly with more than just a few things, and crowd out the important (required) fields. Instead, the table in a collections tab should be editable itself (given the collection owner is the one logged in). Akin to [something like this](http://www.editablegrid.net/en)

I haven't completely been following the entire discussion, so please correct me if I am not interpreting this new metadata that will be added correctly!

vsoch:
We should also be cautious about how we add this metadata to the API. Akin to how the images for a collection are returned with a separate call, the custom metadata should be as well:

http://neurovault.org/api/collections/457/metadata

vsoch:
I just realized, we would be giving programmatic API access to a subset of data without requiring any structure...

my dream is coming true! :wink:

ljchang:
Sorry, I'm a little late to this.  @vsoch any chance you saw the PR with the editable table that @burnash built?  It's pretty slick and I believe is based on <a href=http://handsontable.com/>handontable</a>.  I think the idea for now is to keep this image metadata separate from the collection metadata in the user interface and also the API.  I believe @burnash had the edit tab for the custom metadata set so that you could directly edit this custom metadata if you had permission.

I like this suggestion:
>http://neurovault.org/api/collections/457/metadata

And also this one:
>we can use something simple like color to distinguish.

vsoch:
I haven't tested it out, I will today when I get to campus!

vsoch:
@chrisfilo I followed your instructions to update the VM, but I am still getting this error:

      hstore type not found in the database. please install it from your 'contrib/hstore.sql' file

I found the hstore file and tried to load it manually:

      postgres@localhost:~$ psql -U postgres -f /usr/share/postgresql/9.3/extension/hstore--1.2.sql 
      Use "CREATE EXTENSION hstore" to load this file.

and it of course just tells me to do exactly what you did :)

burnash:
@vsoch did you managed to make it work?

Just make sure you're connected to the `neurovault` database. Depending on your database privileges settings you can even do this under your normal `neurovault` user (if that's the owner of `neurovault` db.)

After executing `create extension hstore;`, you can check if the extesion has been installed correctly with `\dx` command. You session may look like this:

```
$ sudo apt-get install postgresql-contrib
$ psql -U neurovault neurovault
psql (9.4.1)
Type "help" for help.

neurovault=> create extension hstore;
CREATE EXTENSION
neurovault=> \dx
                           List of installed extensions
  Name   | Version |   Schema   |                   Description
---------+---------+------------+--------------------------------------------------
 hstore  | 1.3     | public     | data type for storing sets of (key, value) pairs
 plpgsql | 1.0     | pg_catalog | PL/pgSQL procedural language
(2 rows)

```

vsoch:
Hmm, I'm not able to log in as neurovault:

      (nv_env)vagrant@localhost:~$ psql -U neurovault
      psql: FATAL:  Peer authentication failed for user "neurovault"
      (nv_env)vagrant@localhost:~$ psql -U neurovault:neurovault

I tried both with and without the venv activated, in case that made a difference.

I also have the ~/.pgpass file:

      localhost:5432:neurovault:neurovault:neurovault

When I do the above with postgres user, I can execute the command but I don't see the hstore created. I also tried changing to neurovault user first (su - neurovault) but it tells me there is no password.

burnash:
@vsoch Try this string to log in:

    $ psql -h localhost -U neurovault neurovault

Without `-h` `psql` by default uses [the peer authentication](http://www.postgresql.org/docs/9.3/static/auth-methods.html#AUTH-PEER) which in turn uses your system user's creds to log in.

vsoch:
Almost - now we have a permissions issue:

      neurovault=> create extension hstore; CREATE EXTENSION
      ERROR:  permission denied to create extension "hstore"
      HINT:  Must be superuser to create this extension.
      neurovault->

I remember back in January or February we had a similar permissions issue, and @chrisfilo resolved it. I'm not sure how he did.

burnash:
Ok, that means `neurovault` user doesn't have enough permissions. Try again @chrisfilo's script:

```
sudo su - postgres
psql neurovault
neurovault=# create extension hstore;
```
This time it'll get through.

vsoch:
Success!

Just a quick note - why are we using the filenames of the image as the unique ID for the metadata? I can't seem to easily find them unless I go all the way to the "details" tab, one by one, and that is quite tenuous. I think we should have a button on this same "upload metadata" page that will let the user download an empty csv template with the image filenames. Or we could use the image id as the unique ID instead. Thoughts?

vsoch:
Ack, I can't test it... the button doesn't work for me.

vsoch:
Oh hang on, I bet you I need to re-collect static files. Let me try that.

vsoch:
This is awesome! Holy :cow:, I love it!

A few comments:

- It's hard to distinguish the capitalization of "Subject Id" because the little and big i look the same. We should perhaps just do Subject ID so it's clear they are both caps.
- If the user tries to use the image names from the interface, the underscores are replaced with spaces, and the filenames won't match up (if he/she does not have them handy as I did).
- When we get the message "Wrong value in field "Image Type" we should give them a set of options that are valid.
- I'm also getting "Wrong value format in field "File Name" so I'm not sure what the interface is looks for - the filename with the spaces?

burnash:
@vsoch congratulations! :)

> why are we using the filenames of the image as the unique ID for the metadata?

I was inspired by #163. If I get it right, before the collection is online, filenames are very handy for this purpose. And given that some metadata for a study may exist before the study gets uploaded to NV —filenames are already there.

You got very good points.

1. You're absolutely right about the obscurity of image filename/metadata association in the UI.
2. I like the idea of CSV template. As for using image ids for metadata unique ID I do need some help with consideration of it, because I may miss some possible metadata use cases. 

> It's hard to distinguish the capitalization of "Subject Id" because the little and big i look the same. We should perhaps just do Subject ID so it's clear they are both caps.

Agreed.

> If the user tries to use the image names from the interface, the underscores are replaced with spaces, and the filenames won't match up (if he/she does not have them handy as I did).

That's right. The code relies on matching up with the original filename, as it is in the "file" field on the image details page (but without url path, e.g. Pain_Subject_10_High.nii.gz.)

> When we get the message "Wrong value in field "Image Type" we should give them a set of options that are valid.

True. This field will be replaced with analysis-level, but generally it's good to tell a user about possible choices.

> I'm also getting "Wrong value format in field "File Name" so I'm not sure what the interface is looks for - the filename with the spaces?

It looks for original filename. I was about to address this with showing a sample CSV file (maybe a screenshot of a grid). But you're right this needs a clear description.

Thank you for your feedback. I took a small break from this PR to build a simple neuro-learn webapp prototype. It's almost done, so I'll switch back to the NV metadata support soon.

vsoch:
Haha yes, this is possibly the coolest thing I've done today :)

> I was inspired by #163. If I get it right, before the collection is online, filenames are very handy for this purpose. And given that some metadata for a study may exist before the study gets uploaded to NV —filenames are already there.

I'm glad you like the idea of giving a template - that would allow for us to use whatever modified name that we please, and make the upload process much easier for the user.

> Thank you for your feedback. I took a small break from this PR to build a simple neuro-learn webapp prototype. It's almost done, so I'll switch back to the NV metadata support soon.

That's exciting! Will you share when you finish?

burnash:
@vsoch Sure, I'll be glad to share.

chrisfilo:
Might be of interest: https://docs.djangoproject.com/en/1.8/ref/contrib/postgres/fields/#hstorefield

burnash:
@chrisfilo This is interesting. Thanks for the info. When do you plan to upgrade to 1.8?

chrisfilo:
We were waiting for django polymorphic, but it seems they are ready: https://django-polymorphic.readthedocs.org/en/latest/changelog.html - so soon!

burnash:
@crisfilo I've fixed most of the issues you and @vsoch outlined in the comments ([comment](https://github.com/NeuroVault/NeuroVault/pull/234#issuecomment-98902843), [comment](https://github.com/NeuroVault/NeuroVault/pull/234#issuecomment-105324954) and [comment](https://github.com/NeuroVault/NeuroVault/pull/234#issuecomment-105327496)) and rebased the branch. 

The one unresolved issue is this:

>  cognitive_paradigm field is missing in the tabular interface.

In trying to figure out how to accomplish this, I'd like to clarify if you want the user to be able to upload a csv file with a cognitive_paradigm field.

This field is different from the other "fixed" metadata fields in StatisticMap objects because cognitive_paradigm_cogatlas is not a scalar value but a ForeignKey. If we support csv metadata import for this field, what should be the format for a value in this field?

burnash:
@chrisfilo Is there anything I can do to help get the pull request merged?

chrisfilo:
Sorry for late reply. I was busy dockerizing NeuroVault (which is sadly not yet finished, but I merged that branch anyway). Here's the game plan:
1. Considering our previous discussions map types and analysis levels - please review this PR https://github.com/NeuroVault/NeuroVault/pull/270. I'll merge it as soon as you sign of on it.
2. Please rebase the branch and include enabling hstor in the new docker framework (you can do this by adding an init .sh file in the docker-compose.yml file, see https://registry.hub.docker.com/_/postgres/)
3. I will then go through the PR again - hopefully everything will be good!

PS Concerning your cognitive atlas question - I think we should go with human readable names. BTW @vsoch figured out how to do autocompletion/dropdown inside your tabular environment. This should be useful for map type and cognitive atlas fields.

vsoch:
It is super easy! See here:

https://github.com/vsoch/CogatPheno/blob/master/cogpheno/cogpheno/apps/assessments/templates/edit_questions.html

burnash:
@chrisfilo @vsoch thanks for the reply and for the how-to link. 

Just a little clarification regarding cognitive atlas. Do you mean the user will input a human readable names (e.g `choice task between risky and non-risky options` or `Children&#39;s Communication Checklist` [sic]) into CSV file, then import it, and the server will work out the corresponding  `cog_atlas_id` to insert a value into StatisticMap model?

chrisfilo:
Yes!

vsoch:
Here is what mine looks like with autocomplete - user searches for the term:

dropdown

and then in the view it's [matched to a concept](https://github.com/vsoch/CogatPheno/blob/master/cogpheno/cogpheno/apps/assessments/views.py#L217).

vsoch:
@burnash - for the sign up log in form (I just looked again, it's beautiful by the way) were you planning on adding quick fixes to the forgot password and user profile forms? I think we are ready to merge that branch when we get those final details. My thinking is that the forgot password should be a modal, and the user profile can be as is (on the page) but with the equivalent bootstrap styling. What do you think?

burnash:
@chrisfilo Good. Can I add an index to the `name` field in the db? 
@vsoch It looks very impressive. Great job!

chrisfilo:
Sure.

chrisfilo:
The new map types and analysis_levels have been merged into master!

vsoch:
How did I miss this? Awesome!! Great job @burnash @chrisfilo @tyarkoni @ljchang !

burnash:
@chrisfilo I've finished the last issues.

* Added the cognitive_paradigm to the grid (with autocomplete & fetching data from the server.)
* Updated tests for the cognitive_paradigm field handling.
* Enabled hstor in the docker-compose config.
* Added the analysis_level field (#270) to the fixed fields list.
* Rebased the branch.

Please review the update.

chrisfilo:
excellent! One thing I noticed is that you are relying on third party custom postgres image. It would be safer and more robust if you could "mount" a config file in docker-compose file:

	create_extension.sh docker-entrypoint-initdb.d/create_extension.sh

where create_extension can be copied from here:

https://github.com/aidanlister/postgres-hstore/blob/master/create_extension.sh

burnash:
@chrisfilo I didn't know the preferred way so I decided to go with that image (since it's in the hub and very simply derived from the official image.)

Do you mean a `db` section may look like this?

~~~yaml
db:
  image: postgres
  volumes:
    - create_hstore_extension.sh:/docker-entrypoint-initdb.d/create_hstore_extension.sh:ro
~~~

chrisfilo:
yup :)

burnash:
@chrisfilo It works :)

screen shot 2015-07-28 at 08 08 01

vsoch:
@burnash I'm going to take a look as well, so there will be a little bit more delay for that.  Nice terminal colors, btw :wink:

burnash:
@vsoch Of course! I'll be glad to get your feedback.

vsoch:
ok - apologies for the delay in getting you this feedback, I am a docker-noob.

I think we've made great progress here, and this will be ready for live time with some tweaks to the user interface / experience.  It's a little confusing having the upload button, then downloading a file, then uploading it, then seeing the sheet, and then being told I am missing fields that I didn't know about. I want to suggest the following workflow:

Click upload metadata --> go to sheet in browser with required fields clearly shown --> copy paste or edit values --> click upload button.

And then either it works and there is a nice green success message, or there is a specific error message to fix something. We do not navigate away from the page. For example, error message:

error_example

And success message:

success_message

Some details:

- the table should have autocomplete for cognitive atlas task, modality, and anything else that must be a fixed value. Given that the selection is not made or is made incorrectly, it can dynamically render as red to tell the user of the error
- the error / success messages should be shown in one spot, colored accordingly.

I've done something exactly like this for my CogatPheno repo, so I think you could pretty easily steal that code to get you started! The good news is that this is mostly UI related, so the additional work should not be too much.  The additions that you would want to make are more validations on the back end to return more specific error messages (I made it just for me, so I just return a success or failure / green or red, and I have intuition about what I need to fix).

You've done such great work, and I don't want you to get discouraged! I think this final detail will marry a good user experience with the functionality that we need, and then will be ready for live time. I hope this makes sense... overall really great job :O)

burnash:
@vsoch thanks for the feedback. Before answering in full length I just want to quickly clarify the suggested workflow:

> Click upload metadata → go to sheet in browser with required fields clearly shown → copy paste or 
> edit values → click upload button.

That way the user does not upload the csv file. Is it correct?

vsoch:
correct!

burnash:
I may miss some use cases, however this PR is addressing the issue #163 so the support of csv import was a prerequisite. Initially there were no required fields except the filename (you need it to identify images to attach values to.) 

The use case the csv import was designed for is when the user has the arbitrary study data (in tabular form) and wishes to store it along with imaging data on NeuroVault. If I get it right the csv format has been chosen because of its ubiquitous support by various data manipulation and analysis systems. The samples of such metadata I encountered were quite big and may not be very handy for copy & paste. 
Again I may miss something. @ljchang @chrisfilo can you weigh in?

The addition of fixed fields (those defined on the image model in Python) requested by @chrisfilo and the requirement to handle those fields the same way as arbitrary key-value data made the import process a bit more complex. Now since there are required fixed fields on the image model the user must supply them in the csv file or she/he will get an error (I'm grateful you've spotted the error btw and I agree it is not a good user experience). So we need to either let the import process [to bypass required fixed fields](https://github.com/NeuroVault/NeuroVault/blob/7948cb64fd10510fd1c1efadc6560012914cf627/neurovault/apps/statmaps/views.py#L473) or to add those fields (along with values) to the template next to filenames:

screen shot 2015-07-29 at 00 43 24

The latter case can be confusing for those who already possess a csv with the metadata since they'll have to merge that data with the fixed fields data from the template prior the upload. 

On the other hand, if a user doesn't have a csv metadata file for a collection why would the user want to upload something anyway? If the user wants to mass edit the metadata (or copy & paste something like in your workflow) it's better be done on the edit images metadata page:
screen shot 2015-07-29 at 01 34 37

I guess it is time for everyone involved to get on the same page and to decide which of the use cases are we trying to address in the metadata import.

vsoch:
> I may miss some use cases, however this PR is addressing the issue #163 so the support of csv import was a prerequisite. Initially there were no required fields except the filename (you need it to identify images to attach values to.)

My only concern is user experience. If having the required fields makes the process weird and confusing, then they should not be there.

> The use case the csv import was designed for is when the user has the arbitrary study data (in tabular form) and wishes to store it along with imaging data on NeuroVault. If I get it right the csv format has been chosen because of its ubiquitous support by various data manipulation and analysis systems. The samples of such metadata I encountered were quite big and may not be very handy for copy & paste.

I agree, copy paste can get messy with very large data. However if we look at the data in NeuroVault, Luke is an outlier. I would estimate most collections have around 10ish, and the larger ones have maybe 20-30. Having a simplified UI without uploading csv would still support the majority of users, and maybe even a size of ~6000.

> The addition of fixed fields (those defined on the image model in Python) requested by @chrisfilo and the requirement to handle those fields the same way as arbitrary key-value data made the import process a bit more complex. Now since there are required fixed fields on the image model the user must supply them in the csv file or she/he will get an error (I'm grateful you've spotted the error btw and I agree it is not a good user experience). So we need to either let the import process to bypass required fixed fields or to add those fields (along with values) to the template next to filenames:

I think integrating the fix fields into the custom meta data is problematic and makes the process highly error prone. I don't see why we are doing this.

> The latter case can be confusing for those who already possess a csv with the metadata since they'll have to merge that data with the fixed fields data from the template prior the upload.

I agree, it's confusing and complicated, and not a good user experience. The upload, I think, should not require any fixed fields, and should be able to support either updating or adding new meta data, (either for all images or just a subset). I wasn't participating much in this original conversation, so it could be that there is some rationale for including these fixed fields...

> On the other hand, if a user doesn't have a csv metadata file for a collection why would the user want to upload something anyway? If the user wants to mass edit the metadata (or copy & paste something like in your workflow) it's better be done on the edit images metadata page

Just having these be "two separate things" is too confusing.

> I guess it is time for everyone involved to get on the same page and to decide which of the use cases are we trying to address in the metadata import.

I agree.

Let's update my previous description to give an option for adding csv, for the rare user like Luke! How about the following:

Click upload metadata → go to sheet in browser with required fields clearly shown → copy paste or 
edit values → click upload button.

With a button at the bottom to "upload csv" that will immediately populate the table. And there are no required fields.

chrisfilo:
I'll have a look at this now. @vsoch there is a use case for big collections of single subject maps. This is the main reason why this work even started!

ljchang:
Just to make sure I'm on the same page.  Is the issue with the user experience because users may not necessarily remember or even know to fill out the required fixed fields?  Shouldn't these fields already be populated when users upload their images in the first place?  if so, couldn't these required fields just be matched via the file name with the new csv file?  I thought @chrisfilo's suggestion to merge the handsontable interface with the image meta data was primarily for when a user wants to edit multiple images quickly.  There should never be a scenario where the csv file is added before the images.  For example, I need to change the map-type and image-level fields for both of my larger collections, but I can't do it yet because there is currently no easy or programatic way to update this information.  I don't have the attention span to manually update 6000 images through individually clicking on the edit details.  My understanding is that the reason why these fields are required is because they are already required when uploading the image data.  This new upload csv function was never meant to replace the existing interface.

@vsoch, the two large datasets are just the test collections for this new feature.  They are outliers for the moment, but once this pull request is merged there will be many larger collections that will be added.  For example, in the grant that Tor and I received to fund @burnash, we proposed to include 10-15 existing datasets that we already have access to.  However, we are basically just waiting until this new tabular interface functionality is added and stable.  I suspect that there will be many more datasets that will follow from other groups as I've already spoken to several other labs that agreed to contribute data. 

This feature is necessary for our new image analysis application that @burnash is building.  This will likely be an incentive for many others to also share single subject data to neurovault.  Unfortunately, we have a bit of time pressure to finish our analysis application as our grant is winding down in about 2-3 weeks and we can't complete it until we have the image meta data associated with neurovault collections.  I know you all are busy, but I really appreciate all of the discussion and also everyone's quick feedback.  It would be immensely helpful for us to have this live soon especially without spending too much additional time making drastic changes.  Perhaps a good time to fine tune some of the interface details might be at the neurosynth hackathon in September?  It would be great to be able discuss these things in person and also have some more data on user experiences for various use cases.

ljchang:
Sort of unrelated, but you are going to <em>love</em> @burnash's new search interface for neurovault collections that he has built for our neuro-learn app.  We are looking forward to sharing it soon to get everyone's feedback.

chrisfilo:
Thanks @ljchang and @vsoch In general this looks pretty good to me! I agree that the "Upload CSV" functionality will have limited audience so maybe we should not expose it the way we do right now. In terms of compulsory fields - I would prefer to keep requiring them. Future meta analysis projects will benefit from this :) 

You asked how come come compulsory fields are not filled. This happens when people do bulk upload instead of adding images one by one. We later force them to fill those fields if they want to edit images. Not a perfect solution, but works for now.

Here's a list of more concrete changes:
* There is no way to add an image to a collection anymore (other than bulk upload). Was that button accidentally removed?
* "Reset field" seems to be a confusing term in context of fuzzy matching.
* In the table edit view copy/pasting does not work between columns with dropdown lists.
* Add compulsory columns to the template CSV as suggested by @vsoch 
* Make the error message more explicit and clear when uploading a CSV file that is missing compulsory columns (unless they are already filled).
* Move "Upload image metadata (.csv)" button from collection view to "Edit Images Metadata" view (as an alternative to editing the table online). This will make this feature less confusing for people not doing bulk uploads.
* Could we get error reporting the same way as in CogatPheno? Highlighting cells with errors looks great.
* There is something wrong with fuzzy matching. Column "awesomeness" I used in my example  was fuzzy matched to "Age".
* As @vsoch noticed when importing a CSV with fixed fields (for example analysis level) there is a) no fuzzy matching of values b) no dropdown for allowed values - just an error that something is wrong.

I hope this helps. Thanks for your hard work and all the comments!

burnash:
@chrisfilo Thanks for the feedback, @ljchang thanks for weighing in.

@chrisfilo While I'm gathering the answers for your list, do you mind to describe a workflow as you see it for someone who uploads the csv file? (Considering the presence of fixed required fields)

chrisfilo:
Ok this is how I see it:
1. User goes to "Edit Images Metadata"
2. Clicks "Upload CSV"
3. You display the "importmetadata" page as usual, but be more explicit about required fields. I would suggest adding them to the example as well as under "Requirements". I would also make all this information visible by default (instead of hiding it until user clicks on it).
4. You can should also add the required columns to the template. Encourage using it.
5. When user uploads the CSV file you should check if the required columns are in the uploaded files. If not there should be a clear error message and encouragement to use the template.
6. When a CSV file with the right columns is uploaded, fixed field values should be validated and marked red if incorrect. There should be a dropdown option to select the correct alternative. Fuzzy matching of values might be too complicated to implement.

Does this make sense?

burnash:
@chrisfilo A suggestion. The PR itself and the discussion behind is quite sizable. What if we split it into smaller PRs? Not only this may simplify the reasoning behind the features, it also can simplify the synchronization of branches (and save some time.) There are certainly parts everyone is agreed upon.

What do you think?

chrisfilo:
Sure - how would you like to split it though?

burnash:
This way:

**Part I**. The low level hstore support, edit images metadata page, and updated UI for collection page:

screen shot 2015-07-29 at 04 24 19
 
*The link to the import metadata page will be removed from the menu, as the all import-related code*

**Part II**. The import metadata page.

chrisfilo:
Works for me!

vsoch:
@burnash , if you need to do extensive work on the collections page and want html, I did a lot of work to convert the haml to html:

https://github.com/vsoch/NeuroVault/blob/fix/progressbar/neurovault/apps/statmaps/templates/statmaps/collection_details.html

It's part of a PR to fix the file upload / add a loading spinner, and it's all ready to go minus fixing the zip upload. I didn't get around to it yet because I was making a site for nipy, heh.

tyarkoni:
Convert the haml to html?!?! What is this heresy!

burnash:
@vsoch thanks again for the feedback and for the link to the converted template. The changes to the page were small so I went with the haml.

@vsoch @chrisfilo I branched #309 from this PR. It is the first part. It has hstore support and edit metadata page. Please review it.

I'm closing this PR now.

PR #262, opened 2015-05-13
Collection property fields

jbwexler:


chrisfilo:
I think you forgot to add migrations. Also could you link your github account to travis CI? Tests will run automatically this way.

neurovault/apps/statmaps/migrations/0038_auto_20150514_0032.py
```diff
@@ -0,0 +1,49 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('statmaps', '0037_auto_20150307_0222'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='collection',
+            name='skip_factor',
```
chrisfilo: This way we will delete user data. Please do it in two step way - add a new field first, move data (you have to manually create a migration for this - see past migrations for examples), and only then remove the old field.

Issue #241, opened 2015-05-11
Add page information to the title


chrisfilo:
It should not be the same for every subpage.

PR #264, opened 2015-05-14
Interface Tweaks for NeuroVault

### LINKED ISSUE #239: left/right content padding is uneven

vsoch:
This PR will address a small set of current issues, as well as changes to the interface. The PR includes:

- update to bootstrap 3 to address #240. The only immediate issues [I saw here](http://getbootstrap.com/migration/) were regarding spans --> columns, and the navigation classes. If you look at "What's new" there are substantial classes that we would find useful, including panels, list groups, glyphicons, etc. I didn't want to add extensive new changes to this PR, but we can come back to these.
- Adding a footer to address #246.  I mimicked @tyarkoni 's style and added links to all of our siblings databases.
- Left/Right padding issue has been fixed #239 
- Interface tweaks: I have made the interface wider to support a better UI for some of the visualization. I slightly separated the header from the content, and added a background of a "brain butterfly," which is elegant and subtle, as I remember you liked one of the NeuroVault logo contenders that was elegant. If this is not the style we are going for, let's chat about what we can do. I tried to minimize too much color, but I muchly liked the blue at the top.
- All page titles are now relevant to the pages to address #241 
- The query image in "find_similar" is now scrolling so the user can always see it.

I stopped there to not put too many things into one PR. I am still not quite happy with details on many of the pages, and these can be addressed in future PR. Importantly, pybraincompare has had changes to allow for [custom specification](https://github.com/vsoch/pybraincompare/blob/master/pybraincompare/compare/search.py#L67) of the visualization widths, and will need to be re-installed for the changes to work correctly. The package is updated on pip with these changes.


vsoch:
If I have some time I want to try making the background into a vector so I can make the pattern without a slight blur.

vsoch:
I made it into a vector and then remade the image, but I'm terrible with Inkscape, so it's not significantly better.
bb-sharp3

Vector outline [is here](vbmis.com/bmi/share/neurovault/brain-butterfly-vector.jpg.svg)

vsoch:
I think there are tricks to making really impressive backgrounds, possibly with background textures. I need to learn more about it.

vsoch:
A "barely there" rendering to emphasize clean, simple white (tilt screen to see). Although it could be an issue to not have discrimination between content box and background.

barelythere

multiple-bb-texture-light-big

chrisfilo:
Shouldn't number of images column be narrower? Have you tried it with more
populated database?

vsoch:
Good catch. It's because the table on the front page uses the old span args - I will update to col-md.

vsoch:
fixed

vsoch:
favicon

vsoch:
That's strange - I didn't change anything...

vsoch:
It's having trouble downloading the [feat data](https://github.com/NeuroVault/neurovault_data/tree/master/FEAT_testdata). I'm not sure why - I'll try restarting it later.

chrisfilo:
I'm not convinced about using inverted coronal slice as a logo/ background.
I would leave logo decision for another PR.

vsoch:
That looks much better :O)
muchbetter

There are also different derivations with spacing we could do:

header1

This is the current:
header2

chrisfilo:
I like the "current". Resembles most material design!

vsoch:
material_design

vsoch:
Needs more space
morespace

chrisfilo:
What do you mean?

vsoch:
The distance between the "NeuroVault" and the header and content below it. In the first picture it's very small (crowded) and the image directly above is more open. My thinking for this is that eventually, we can have that space be more of a content slider (for example, it would start on "NeuroVault" and then transition to "map of the day" or something like that.

vsoch:
In other words, we want him to be fat *and* tall :)

chrisfilo:
could you also update "bootstrap.min.js"?

chrisfilo:
there is something wrong when there are more than one page
screen shot 2015-05-22 at 10 05 10

Also I'm not sure how to feel about mixing violet with blue.

vsoch:
Where is violet?

chrisfilo:
Could you cherry pick be39d39 and send it as a separate PR so other tests will stop failing? Thx.

chrisfilo:
violet/purple - used in odd rows

vsoch:
Didn't we do that fix with PR https://github.com/NeuroVault/NeuroVault/pull/267? The [current requirements.txt](https://github.com/NeuroVault/NeuroVault/blob/master/requirements.txt) has:

     git+https://github.com/incf-nidash/nidmresults.git@5a9c494918de8e1c05a981e40a2a65fc4a4189be

Do you mean to try and get it working with the current nidm results?

vsoch:
I think we just found "the NeuroVault" dress - I don't see violet in the rows in the picture above!

chrisfilo:
* Yes you are right about the nidm commit.

* Well - the odd rows are not blue and it's not grey...

screen shot 2015-05-22 at 10 17 01

"This collection is empty" text used to have bigger top padding. Links below used to be buttons.

* Please remove "copyright" from the footer. It's meaningless in our context.

* Please center papaya viewer or make it bigger.

vsoch:
thanks for the feedback :)

vsoch:
I have made the following changes to the PR:

- the footer has the "NeuroVault" copyright removed, and is spaced slightly lower from the context
- the pagination for all tables is fixed. The issue was that these templates are defined with jquery datatables, and I had to update this and add a bootstrap styling and .js file specific for the datatables for it to work with bootstrap 3.
- The buttons shown for the empty collection are restored. I had to add .btn-primary class.
- the Atlas comparison page had a bug - the font was too small (11 point) and this is fixed.
- I have added some top padding below the main navigation buttons. This was likely changed do to bootstrap upgrade.
- The papaya viewer is now page width.
- I tried really hard to find purple, but I did not. Here is how the tables look to me:

nopurple

I am including the purple palette to the left for comparison.  Specifically, here is the color used for the rows:

blue

It is a shade of light blue! It matches the site.

vsoch:
Looks like the font size is a bug in the current deployment too:

http://neurovault.org/images/compare/26/27

oups. :blush:

chrisfilo:
thanks for the updates. Here are some more things I found:

screen shot 2015-06-01 at 16 45 47
* why is there a dot before "images in collection?
* "add a new image" should be a button
* it would be great if you could make input form wider


* when I try to "upload folder with images" the screen greys out and nothing happens


screen shot 2015-06-01 at 16 50 29
* what are those squares?


screen shot 2015-06-01 at 16 51 34
* those radio buttons should be aligned with the page

vsoch:
I'm having some trouble figuring out the styling of the forms. For example, I want to add a helper field to [style the columns](https://django-crispy-forms.readthedocs.org/en/latest/crispy_tag_forms.html#bootstrap3-horizontal-forms) as bootstrap "col-md-5" with something like:

      helper.label_class = 'col-lg-2'
      helper.field_class = 'col-lg-8'

The "edit images" page (I think) uses a [CollectionInlineFormSet](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/forms.py#L569) via a [CollectionFormSet](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/forms.py#L616)

So I added the helper to the [PolymorphicFormSet init](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/forms.py#L505) but it didn't seem to change anything in the page. I think it might have to do with how the form is rendered in the [template](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/templates/statmaps/edit_images.html#L91). Any pointers?

vsoch:
I'm also having trouble debugging why clicking "upload folder" goes gray and nothing happens - I didn't make any changes to the function. Did @jbwexler add the upload progress bar? We could talk about how it works to figure this out - it doesn't make sense that changes to the interface would break it.

vsoch:
woot! Figured out the forms problem :)

vsoch:
[this guy](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/settings.py#L241)

chrisfilo:
ah! well done!

vsoch:
And I think this might be (one) issue with the folder upload - will look into tomorrow:

>> Modal markup has changed significantly. The .modal-header, .modal-body, and .modal-footer sections are now wrapped in .modal-content and .modal-dialog for better mobile styling and behavior. Also, you should no longer apply .hide to .modal in your markup.

vsoch:
Ok, I've fixed all these issues! I'm looking through the pages again, want to do some tweaks to the new FAQ, and I should have an update to the PR within the next hour.

vsoch:
This update addresses the following:

### Edit images form
 - The form was not being styled, and it was due to the fact that we needed to set the template to be bootstrap 3 in [settings.py](https://github.com/vsoch/NeuroVault/blob/master/neurovault/settings.py#L241). This resolved the input width issue, as well as the general ugliness. 
- The <ul> list on the left side of the "edit images" page was also rendering ugly because there was a gray border around the top element, but not the remaining in the list. I removed this bordering to give it a cleaner look:

editimages

The above also shows that the <ul> has been changed to a header, making it slightly larger to fit in the space, and the dropdown at the bottom has been fixed (to again be a dropdown) and spaced to line up with the left column and centered vertically.

### Single edit image page
I noticed that our "submit" buttons on the edit image pages are not specified to be large, and wanted to check that this is intentional.

### Edit collection page
There were several bugs on this page.
- First, the buttons were too small. This was due to the fact that the "btn-large" class was changed to "btn-lg" for [bootstrap 3](http://getbootstrap.com/migration/). I fixed this. 
- Weird squares: This was an issue present across tables, and I didn't catch it because the squares did not appear in the browser that I was first testing on Ubuntu. The issue was that the datatables styling for boostrap 3 including a set of glyphicons in a "fonts" folder that we did not have. I added these to our repo, and instead of using the datatables.boostrap 3 css file from a CDN, added a local version to delete unnecessary symbols, period. It now looks like this:

collection

Upload folder with images: was just not working, and this was due to pretty extensive changes for the "modal" object in bootstrap 3. I was not familiar with modal, and spent most of the morning translating the .haml page to .html because I was having trouble testing with .haml (I'm just not as familiar with it). The bug came down to two things: 1) This line:

      $("#upload_process").on "shown", (e) ->

Needed to be updated to:

      $("#upload_process").on "shown.bs.modal", (e) ->

And 2) the modal div itself needed a .modal-dialog and.modal-content added. Likely it would have been fine without this, but having these top divs was suggested, so I added them.
- Finally, the upload by folder did not work for an empty collection (this possibly was a bootstrap 3 bug?), and I thought that this was due to the fact that the file uploader div was only present in the top if statement. I [added it to the else](https://github.com/vsoch/NeuroVault/blob/update/interface_tweaks/neurovault/apps/statmaps/templates/statmaps/collection_details.html.haml#L184) to fix the problem for my repo.

### Radio buttons
This was another tricky one because the buttons are being rendered by crispy forms. To fix this spacing, I [edited the style](https://github.com/vsoch/NeuroVault/blob/update/interface_tweaks/neurovault/apps/main/static/css/style.css#L211).

chrisfilo:
Looks much better, but I found two issues:
* I cannot add images through "one by one method" for empty and non-empty collections
* cognitive paradigm field is not rendered using select2 widget and thus is not searchable
* heading of the form does not update when you type in a new name in the form (as it used to)

I restarted nginx, uwsgi and collected static files.

vsoch:
ok, I'll check out those points! Thanks for looking at this so promptly!

vsoch:
### Heading Form Update
The javascript was [hiding here](http://neurovault.org/static/scripts/clone-form.js), and the issue was selecting the element "control-group" instead of "form-group" (change for boostrap 3). It's amazing the kind of things that are unleashed when there is a decision to change something as trivial as a name...

### Select2
Also needed tweaking for boostrap 3. I just totally missed this.

### Base.js
Any reason that we still have [this file?](http://neurovault.org/static/scripts/base.js)

### Submit form
To add detail to this bug - a new image is not added given that other required fields for images that exist are empty - when I have fields missing, the new image disappears and I see red text indicating for me to fill in the values. When the fields are all filled in, the image is added successfully. This is the last but, and I can't figure it out.  Do you have any insight?

chrisfilo:
Did you commit/push everything? Heading, select2, editimages still do not work on my machine. I did collectstatic, restarted servers and cleaned the cache.

vsoch:
Oops, edited the files in the neurovault/static only... one sec!

vsoch:
Update to the bug: after clicking "Save" I still see the new image in the HTML:

imagehtml

but not on the page:

page

This is weird!

chrisfilo:
Interesting... I see something similar in production recently. After the server is running for a few hours editimages stops saving maps:
- it only happens when collection is empty
- it happens even if all fields are filled without errors
- it goes away if I reload the uwsgi server

I think those two are related. Maybe fixing the bug in the branch from this PR will solve the production problems. That would make me sleep at night again...

vsoch:
If that's all it takes to get you sleeping again!! I will work more on this!

vsoch:
Woot! I just found the bug for this particular error, but I'm not sure if it's related to the uwsgi bug you noted. The issue was that the form was submitting as clean, but it clearly wasn't. It came down, again, to looking at the form classes. Specifically, I compared the currently live (old bootstrap) vs the PR (bootstrap 3) and noticed these differences in the control-group/form-group classes:

       # old bootstrap
      class="control-group error"

       # bootstrap 3
       class="form-group has-error"

So then I searched for error in the javascript and found this in the formIsClean function:

       if($(sele).find('div.errors').length === 0 && 
        $(sele).find('div.error').length === 0 &&
        $(sele).find('div.alert-error').length === 0) {
         return true;

and that explains why it returns "True" when the form isn't clean, submitting the form but it doesn't add the image because it's not valid. I changed it to:

      if($(sele).find('div.has-errors').length === 0 && 
        $(sele).find('div.has-error').length === 0 &&
        $(sele).find('div.alert-error').length === 0) 
      
And the form works as it did before! Although I don't think this fix will help you get more sleep :/

tyarkoni:
Hey, so the odd row color is purple in my book too, and it looks pretty bad to me against the dark blue. Can we change it to something lighter and bluer?

vsoch:
Am I color blind? It seriously is blue...

tyarkoni:
Well I'm less concerned with the labeling than with the fact it clashes pretty badly (to my eye, at least) with the navbar and buttons. Maybe other folks can weigh in. @burnash? @ljchang?

vsoch:
And as a reminder, the HTML color picker calls it blue :)

blue

tyarkoni:
Well in CMYK, you get it by [blending cyan and magenta](http://www.colorhexa.com/e2e4ff), so... purple, I say! 

But seriously, it doesn't really matter whether it's blue or purple. My concern is that it looks ugly. If it looks good to everyone else, that's fine, but I suspect at least @chrisfilo shares my aesthetic sensibilities. :p

vsoch:
If it's really purple... I feel like a mole rat. :see_no_evil:

vsoch:
but also note that the ENTIRE blue spectrum is between cyan and magenta... so I'm not convinced that defines it as purple :) Unless there is no blue in the Talverse?

tyarkoni:
For realz though... I'm 100% fine calling it blue. Sapir-Whorf be damned, the tables look equally ugly to me whether we call it blue or purple. It's the ugliness I care about, not the blueness.

vsoch:
Poor NeuroVault is an ugly mother-trucker, but we love him anyway!

And for realz, I'm happy changing the color to any hex code to your heart's content! I shouldn't be trusted to choose it though :)

tyarkoni:
See, I was worried you were going to say that. And I'm not prepared to handle that kind of responsibility. So I'm going to insist that you choose a color.

Oh, fine: try #DFEFFF.

p.s. NeuroVault was way uglier back when I was on the job

burnash:
I agree with Tal. On my screen the odd rows look very intense. Especially when [f.lux](https://justgetflux.com/) kicks in.

Why not to go with a default bootstrap's color for [striped rows](http://getbootstrap.com/css/#tables-striped)? It looks very clean, does the job of distinguishing rows and there's no hue to argue upon.

E2E4FF:

blurple

Bootstrap:

gray

vsoch:
We figured this out! It looks blue on JUST my computer, and @chrisfilo confirms! So the good news is that I'm not color blind, and that we will definitely do a fix to this.

vsoch:
@burnash if you are suggesting #E2E4FF that implies they are being rendered as a different color? What color is being rendered and where is it defined? For me - the current tables (for example, a datatable on the collections page) are already #E2E4FF, and it's defined in the datatables css:

e2eff

and the table on the main page (that I remember looked purple on Chris screen) is actually not a datatable, but just defined by bootstrap, and the row color is "f9f9f9":

f9f9f9f9

and the table class is "table-striped"

So it appears we have a mystery of two blues! Does everyone else see that?

tyarkoni:
I think @burnash was suggesting--and I would agree--just getting rid of the DataTables colors and relying on the default bootstrap for striped rows. It's possible to have DataTables just use the bootstrap styling; there's a separate plugin for that. See https://www.datatables.net/manual/styling/bootstrap.

tyarkoni:
Oh, and as the comment on that page says, if you just drop the datatables css from the asset pipeline, that might actually be sufficient without having to do anything else (because I think I already added the DT bootstrap libs at one point, though I could be wrong).

vsoch:
But isn't the issue not with the datatables.css - but with the tables on the first page that have rows colored with #f9f9f9? Could you verify which tables look purple to you? The tables that are styled by datatables boostrap already are the "correct" color,  #E2E4FF. If I'm totally missing something, then maybe someone else can do the edits to resolve this issue.

tyarkoni:
I don't follow. the purple rows are #e2e4ff, which we want to get rid of. The front page looks fine as-is; the #f9f9f9 looks good. It's the collections and other pages that are using the datatables color and look bad. We simply want the DataTables to stop using the default DataTables styling and use the default bootstrap styling.

burnash:
@vsoch Sorry for the ambiguity. @tyarkoni's is right. What I suggest is to get rid of datatables color and to default to bootstrap. That's easy with Tal's findings. I've tried to just drop the datatables.css from collections page and it works, but has one defect—the sorting icons in columns are gone:

screen shot 2015-07-06 at 23 47 02

But if you replace the datatables.css with a [plugin file](//cdn.datatables.net/plug-ins/1.10.7/integration/bootstrap/3/dataTables.bootstrap.css), the icons are back (but in a different flavor):
 
screen shot 2015-07-07 at 00 12 46

I've tested this in devtools in Chrome and didn't find any other differences.

PR #268, opened 2015-05-21
Use pip to install nidmfsl package

cmaumet:
Following the merge of incf-nidash/nidm-results_fsl#35, Neurovault can now depend directly on the master branch of both [incf-nidash/nidmresults](https://github.com/incf-nidash/nidmresults) and [incf-nidash/nidm-results_fsl](https://github.com/incf-nidash/nidm-results_fsl).

From the [test report](https://travis-ci.org/cmaumet/NeuroVault/builds/63491857), I think that the NIDM export is going through but the test is failing [here](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/tests/test_feat.py#L104). Do you have any idea of what might be causing this error?

Also, it looks like that there is no `feat3_stats` file in the `log` directory of FSL's feat analyses. This means that nidmresults do not export the noise FWHM (which is fine but could probably be improved). Can you let me know what version of FSL you are using to run feat?

Thank you!

### DIFF
--- a/neurovault/apps/statmaps/tests/test_feat.py
+++ b/neurovault/apps/statmaps/tests/test_feat.py
@@ -23,7 +23,7 @@ class FeatDirectoryTest(TestCase):
                 'fileuri':'ds105.feat.zip?raw=true',
                 'num_statmaps':2,
                 'export_dir':'ds105.feat/cope1.feat/nidm',
-                'ttl_fsize': 37872,
+                'ttl_fsize': 39595,
                 'map_types': ['T','Z'],
                 'names':['Statistic Map: group mean', 'Z-Statistic Map: group mean'],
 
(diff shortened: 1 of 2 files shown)


vsoch:
Thanks Camille! I think I'm useless to answer your question (I don't have access to production server) but I just wanted to say thanks for fixing this :)

cmaumet:
:)

chrisfilo:
The tests are failing because the size of the generate turtle file has changed. If this is ok, please update the expected turtle file size here: https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/tests/test_feat.py#L26

I'll get back to you on the FSL related question.

chrisfilo:
Version 5.0.8 was used most likely (however there is a small chances 5.0.7 was used). Oh the need for provenance...

chrisfilo:
Please reabase/merge to make tests pass (we fixed this with another PR).

cmaumet:
@chrisfilo: I have just rebase but apparently the branch is already up-to-date.

cmaumet:
I have just updated `requirements.txt` to use the `nidmfsl` package available through `pip install` rather relying directly on Github. 

The error is still the same so I think the pip install works correctly.

PR #278, opened 2015-06-14
Restyling the Log In page (and a custom password reset page)

burnash:
Congratulations on migrating to BS3.

I've decided to tune the login page a little bit and to add a missing "forgot password" link:

login-page

The code for the form is a slightly modified version of @amatellanes's [login bootstrap snippet](https://github.com/amatellanes/bootstrap-snippets/blob/master/aboutme_login_style.html). 

I've replaced Zocial with BS buttons and Font Awesome icons, since they're already being used and may better fit the new visual style.


vsoch:
Ah that looks so good!

vsoch:
What do you think about:

- eliminating the "Log in or sign up" at the top (the log in part is implied)
- add a small link to "Sign up" in bottom right of the login form?
- the space above the social links would be empty / clean / white

for a more simplified and clean UI?

vsoch:
This also makes me think - we should have a nice dashboard for users right after they log in. Even something simple like their username, recent contributions / links to images or collections, or a "randomly" chosen message (encouraging to tag with contrasts, add a piece of meta into, etc) to greet them.

burnash:
I agree with you that eliminating the "Log in or Sign Up" at the top makes the form cleaner. This is how it may look like (along with "Sign Up" link at the bottom right:

nv-sign-up-bottom-right

The problem I see with this layout is that "Sign Up" link can be easily left unnoticed. This is a quite important link for the newcomers who like to register with a username/password. But that's the only place it appears on the website.

I found that some sites with a similar Sign In page layout use tab panels to make the form navigation clear:

so-auth

soundcloud-auth

My suggestion is to use the same technique and to restyle the Sign Up page as well. The Log In / Sign Up form may look like this:

nv-login

nv-signup

The tabs provide additional "You are here" clue to the user.

vsoch:
I have some comments - I'm out and about on my :bike: so expect to hear from me later this morning!

vsoch:
Ok, so this comes down to a balance between aesthetic and usability. Here are the questions that I ask myself when I look at this design:

- how often does the user need to find the element (sign up link)
- and is it easy to find?

And the answers:

- Only once, and then never again
- It's pretty easy in both cases

Given the clean UI, I'm not convinced that the Tab is significantly easier to find than the subtle text in the bottom right. In fact, I would say that given that the user knows he/she is on a "login" page, it is expected to find a "sign up" link in a more subtle position. You selected examples to show with tabs, and I could show you an equivalent number of sites that make the "sign up" link a lot more subtle:

subtle1

and this is reddit

reddit

Why do/can they do this? Because the user that wants to sign up has a high incentive to find that link. They are looking for it, and it's expected to be more subtle (and I would go as far to say that having "sign up" in the bottom right isn't nearly as hard to find as some of these other examples.) Further, the reason that this strategy is ideal is because signing up is something that we do once, and never again. Tabs make sense to separate content that will be equally browsed, consistently over time. After the user has an account, that second tab is just clutter.

But surprisingly, I found significantly more sites that have absolutely no links to sign up from the login page. Why is that? Because if a user comes to a "login" page looking to sign up, we have probably done something wrong. The sign up link should be clear and obvious for anyone not logged in on the main site navigation bar:

yelp

signuplogin

github

If anything, it's a courtesy to have a "sign up" link on the login page form.  A user who wants to sign up will not know to click "Login" at the top right in the first place. Signing up via the login page is not totally intuitive, and it's one click too many. So the answer is not to fiddle over how salient the "sign up" is on the login page, but to add a more clear "sign up" option from the getgo, in the main navigation when the user is not logged in.

Finally, the reason that the Tabs is aesthetically more pleasing is not due to the Tabs themselves, but to the extra space that they add at the top. This spacing should mirror the NeuroVault text on the front page. For the example you showed, here it is with more space (apologies this is a little hacky, I am not great with Gimp):

a28e0f04-12bd-11e5-82b0-44224440cc02

In summary, here is my suggestion / opinion:

- Use the simplest version of the form (sans tabs) above, with padding added to the top and bottom.
- Add a "sign up" link in the navigation (given not logged in) to go to the Sign up page.
- On the login page (above) change the "log in" in the top right to "Sign Up" to offer yet another avenue for finding it.

Overall, great work! This new page is totally beautiful, and these are just details.

burnash:
Totally agree with you that the better option is to add a "Sign Up" to the nav for the unauthenticated page view. I skipped this option since I tried to impact as less of the UI as possible.

From this "scoped edits" prospective tabs were the obvious choice: they both provide navigation options and serve as each section's title. Still I agree with you, we can remove the tabs if we provide another avenue to find "Log In" or "Sign Up" pages.

However I believe that every page or section should have a clear caption. Even in obvious situations.

Take a look at the examples you've given:

bee585c0-1280-11e5-95a4-710c6d5d3fb4

screen shot 2015-06-15 at 01 16 55

Reddit:

screen shot 2015-06-15 at 01 17 47

screen shot 2015-06-15 at 01 18 26

screen shot 2015-06-15 at 01 18 41

GitHub has modest Sign In:

screen shot 2015-06-15 at 01 19 40

But rich Sign Up:

screen shot 2015-06-15 at 01 27 00

What are your thoughts on how the NeuroVault's Sign Up page should look like?

vsoch:
When you say "rich sign up" you mean there are lots of things to read, tabs, etc? I do like the idea of having "sign up" be distinct from Login (for example, the pop up form is elegant), but is there a reason to make it more complicated than that?

burnash:
No, I'm not advocating any complex page with lots of things to read. That was just a sample of two distinct pages.

Can you elaborate on the pop up form and how you see the difference between Login and Sign Up? 

To summarize: I have the html for the Sign Up form ready. I agree to remove tabs, but without them these two forms look very similar. My proposal is to add captions. What do you think about it?

vsoch:
I can put together an example of what I have in mind sometime this week.

vsoch:
I'm testing this live now - I pulled your branch, collected static, and restarted the server, and it looks like this:

missing

Did you add all your files to the repo from neurovault/static?

vsoch:
Fixed it! Needed to clear my browser cache.

vsoch:
I just submit a PR for you to take a look at the "pop up" that I had in mind.  More details:

https://github.com/burnash/NeuroVault/pull/1

burnash:
Great, thank you! I'll take a look!

vsoch:
ok, more suggestions here!  https://github.com/burnash/NeuroVault/pull/2
Thanks @burnash, this working together thing is pretty fun :O)

chrisfilo:
As far as I am concerned all of the options look much better than the current one. Are we ready to merge?

vsoch:
If you follow the conversation, @burnash was just fixing up the password reset forms, and then everything should be good to go! We decided that the user profile was outside of the scope of this PR  and we would think more carefully about how the user login, and reset forms are styled for another round. Check with burnash before merging because we were working on another branch, and it would be good to verify that all changes have been compiled here. Also verify that everything is functioning as it should. Can't wait for you to see it live!! :)

burnash:
@chrisfilo, @vsoch is correct. I have [another branch](https://github.com/burnash/NeuroVault/tree/restyle-sign-in-sign-up) descended from this one which contains @vsoch's thoughtful contributions and a restyled sign up form. The discussions and some screenshots are here: https://github.com/burnash/NeuroVault/pull/1 https://github.com/burnash/NeuroVault/pull/2

The remaining work as discussed is the following:

1. Decide on how to style the pages in the password reset workflow. In this branch I styled it using the same layout as in log in / sign up forms:
forgotpw
2. Add a missing template `registration/password_reset_confirm.html` where the user actually types in a new password. (I forgot to add this template in the original PR.)
3. Style the "edit profile" and "new password" pages. For both pages we just style the form with bs3, and have it oriented to the left of the page.

Point 1 & 2 can be done very fast if we use same the layout as on the screenshot (possibly with minor css tweaks) because the html is ready.
Point 3 could require a bit more work since I haven't touched these forms.

I'm even open to the idea of splitting the PR in two parts (1 & 2; 3).

vsoch:
@burnash - I should be back and able to work at a computer tomorrow, so if you have design choices you want feedback on, I can offer! Let me know.

vsoch:
Moving this to the correct PR...

@burnash - is the branch "restyle-sign-in-sign-up" the most up to date? Or should I look at this one? 

Given the sign-in-sign-up work -  (I just looked again, it's beautiful by the way) were you planning on adding quick fixes to the forgot password and user profile forms? I think we are ready to merge that branch when we get those final details. My thinking is that the forgot password should be a modal, and the user profile can be as is (on the page) but with the equivalent bootstrap styling. What do you think?

burnash:
@vsoch Yes, the restyle-sign-in-sign-up is the most up to date and this PR is obsolete. I'm closing it now and will resubmit the proper branch. I plan to finish the fixes of the forgot password & user profile pages after I finish with the metadata PR fixes. 

I don't think that making the forgot password a modal is practical. I outlined my reasoning [in the fork's comments](https://github.com/burnash/NeuroVault/pull/2#issuecomment-113968390).

vsoch:
Sounds good! I missed that particular comment, and I agree. Looking forward to getting this merged soon!

Issue #269, opened 2015-05-22
Update NeuroVault to use pybraincompare transformation functions


vsoch:
We want NeuroVault and NeuroSynth to have synchronized calculation of image transformations (even if saving methods are not equivalent). See https://github.com/vsoch/pybraincompare/issues/13

PR #284, opened 2015-06-17
updating tasks to use pybraincompare transformations

vsoch:
This PR will update the function "save_resampled_transformation_single" to use the [same function in pybraincompare](https://github.com/vsoch/pybraincompare/blob/master/pybraincompare/mr/transformation.py#L14) to produce an reduced transformation image vector.  As we talked about https://github.com/NeuroVault/NeuroVault/issues/269, this will allow for a common function to use between NeuroVault and NeuroSynth. The functionality is identical, it's just moving the code from NeuroVault into pybraincompare. To test, I compared a new transformation using the pbc function with an old transformation of the same image:

      import numpy as np
      vec1=np.load("transform_4mm_458.npy")
      vec2=np.load("transform_4mm_458_old.npy")
      (vec1[-np.isnan(vec1)] == vec2[-np.isnan(vec2)]).all()
      True

Note that numpy does not consider nans to be equal (this image has 3 or 4) and this is why I removed them before comparison. I tested this manually in the manage.py shell, as well as automatically with celery running and uploading new images.

### DIFF
--- a/neurovault/apps/statmaps/tasks.py
+++ b/neurovault/apps/statmaps/tasks.py
@@ -1,20 +1,22 @@
 from __future__ import absolute_import
-import os
-import numpy
-import pylab as plt
-import nibabel as nib
+from pybraincompare.compare.maths import calculate_correlation, calculate_pairwise_correlation
+from pybraincompare.compare.mrutils import resample_images_ref, make_binary_deletion_mask, make_binary_deletion_vector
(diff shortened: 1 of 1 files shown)


chrisfilo:
Why did you close it?

vsoch:
If it's decided to be salient it can always be re-opened.

chrisfilo:
Merged via 46f342482269d4d7954de25eb8d7d6a5435f07c5

PR #304, opened 2015-07-19
Fix/progressbar

vsoch:
This PR will add a loading spinner to display when the user is uploading files. Specifically, I did the following:

- got rid of the mother-trucking yaml, replaced with html
- designed a custom loader with neurons (I do not predict this will be appreciated :) )
- I grouped the options to upload images into a single button - this has been bothering me for a while because it's not well organized, and confusing to have three buttons on the same level to do the same thing. It's one extra click to see the dropdown options, but I am in favor of having a clean and organized UI for this click. 
- created two versions - [the current PR branch](https://github.com/vsoch/NeuroVault/tree/fix/progressbar) (with uploading a folder)
and I also created a classic file uploader via a modal ([see this branch](https://github.com/vsoch/NeuroVault/tree/fix/progressbar-modal)). I don't think the functionality is better here, but wanted to give it as an option. This was a new branch that I was testing, and I was able to figure out the fix here, and then apply to the upload_folder option. It's yet another click, and would only be desired if we want the user to be able to select individual files instead of a folder. My guess is no :)

This bug was a stinky one, so stinky! :bug:


neurovault/apps/statmaps/templates/statmaps/collection_details.html
```diff
@@ -0,0 +1,305 @@
...
+                {% endif %}      
+                {% if edit_permission %}
+                    <a class='btn btn-primary' href='/collections/{{ collection.id }}/edit'>Edit collection</a>
+                    {% if collection.image_set.all %}
+                    <div class="btn-group">
+                        <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                        Add images <span class="caret"></span>
+                        </button>
+                        <ul class="dropdown-menu">
+                            <li><a href='/collections/{{ collection.id }}/addimage'>Add single image</a></li>
```
chrisfilo: "Add a single image"?

chrisfilo:
I'm afraid the whirling neurons are too edgy for my taste.

Modal is pretty cool. I wonder if it would make sense to have both folder upload and modal. What do you think?

vsoch:
I will change the animation to a more standard one, I anticipated that it was a bit much, hehe. Edgy is a nice way to describe it, I'd call it something that got spat out of an amalgamation of Kid Pix and a Tellitubbies episode, which can be best described as "ohno!" :)

I am definitely a proponent of having an option for a standard file upload. The issue with only having a folder upload is that it doesn't give any control over selecting a subset of files. For my own usage, I never have used the archive (.zip) upload, and adding images one by one is sort of painful, and that leaves folder upload. But I don't always want all of the files in a folder, so I would have to generate temporary folders, and that was annoying. Thus, we should have the file selector modal as an option! 

I will make this modification with an edit to the current PR - time-frame is likely toward the end of today. But you're floating around somewhere in Europe anyway, so probability says you'll be sleeping anyway! So we can say, tomorrow! :O)

chrisfilo:
Please mind that folder upload and probably modus as well will not work on all browsers (and those options should not be even shown on them). The zip upload is a fallback for bulk uploads using things like IE.

vsoch:
Will do.

vsoch:
ok, should be ready for a look.

neurovault/apps/statmaps/templates/statmaps/collection_details.html
```diff
@@ -0,0 +1,364 @@
...
+                    <div class="btn-group">
+                        <button type="button" class="btn btn-primary dropdown-toggle" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                        Add images <span class="caret"></span>
+                        </button>
+                        <ul class="dropdown-menu">
+                            <li><a href='/collections/{{ collection.id }}/addimage'>Add a single image</a></li>
+                            <!--[if !IE]> -->
+                            <li><a id="show_modal">Add multiple images</a></li>
+                            <!-- <![endif]-->
+                            <li><a id='upload_archive' href="#">Upload archive (.zip)</a></li>
```
chrisfilo: "an archive"?

neurovault/apps/statmaps/templates/statmaps/collection_details.html
```diff
@@ -0,0 +1,364 @@
...
+                            <!--[if !IE]> -->
+                            <li><a id='upload_folder'>Upload folder</a>
```
chrisfilo: "a folder"?

chrisfilo:
Looks great! One thing - when I cancel the select file dialog the spinning wheel does not go away. Could you possibly grey the page out after user selects a file/folder?

Steps to replicate:
1. Press upload folder
2. Press cancel in the dialog box

vsoch:
ok, so the cleanest way to get this working is to turn the upload folder into a similar format as the upload files - meaning having the submission controlled by form submission, and that way the cancel button merely puts the user back to modal without any files submitted. It again is an extra click, but it should work. Does this seem reasonable?

vsoch:
Or we can rid of upload folder entirely. It's a bit redundant given that the user can upload files (control + A is equivalent to "upload folder") and zip.

chrisfilo:
We need upload folder for FEAT output parsing (you can upload a whole FEAT
folder and it will be converted to NIDM-Result) as well as recursive
uploads.

I'll have a look at this soon!

BTW does modal work on IE?

vsoch:
The FormData() doesn't work up until version 10, and since that's the bread and butter of our upload (for folder and files) we disable it for both, meaning also that there is no modal functionality for IE users. However, I think modals would work as long as we don't use the "fade" class, but I don't have IE to test this. I think I'll add a warning (equivalent to the thresholded maps styling) for IE users about the upload. They should be using a better browser anyway :)

I'm going to try to duo-modal thing for the folder upload as well, and I'll think about this on my walk to campus. It's sort of redundant, but if it works that's what is most important.

vsoch:
Actually, I think we should change the message "empty folder" to "no files selected" that seems more standard.

chrisfilo:
Couple of things:
* upload .zip does not work (could we get a test for this?)
* hovering over "add multiple images" and "upload folder" changes cursor to "select" instead of "link"

vsoch:
Yep.  Are these examples of zips that should work?

https://github.com/NeuroVault/neurovault_data/tree/master/FEAT_testdata

chrisfilo:
any zip with multiple .nii.gz files in it should work

vsoch:
Heads up! Just got it working :) I just want to do some dynamic changing of some of the input buttons so it at least (looks) like they are different modals (they are exactly the same :) ) 

I can probably push the fix tonight, but the test will have to come later because I'm sleepy!

vsoch:
hmm looks like it wasn't up to date with neurovault master (upstream was chrisfilo master!) Let me see if I can merge quickly.

vsoch:
ok, now it seems correct.  

I decided to have one function for uploading files, and it calls the zip upload function in the case of finding a zip, but after checking first if it is nidm. I want to check that [this line](https://github.com/vsoch/NeuroVault/blob/fix/progressbar/neurovault/apps/statmaps/views.py#L360)  should be an elif, meaning that the actual upload of the nidm files is taken care of in the first if. 

Having one functions means that the interface calls the same function, with a slight tweak to the title of the modal so it seems different. The benefit is that a user can select one or more files or zips in either case, and it will work for either, and the differentiation between the upload zip and upload files button is really just for the user to know that there are both options.

The only reason to separate the upload_zip function from the monstrous upload_files is to make it more humanly readable.  In time we could probably do the same thing to the rest of the massive function, but it's not necessary.  Changing the name of the function to "upload files" means that "upload_folder" to "upload_files" is changed in url.py, as well as the collection_details.html template. I point this out because if there are other instances of this, we would need to make the change there as well. 

The major stuffs are working, please let me know if there are final tweaks! I think it's pretty reasonable I could do this at sometime in the middle of the night when some catastrophe at Lake Lagunita wakes me up with a cell phone message, or first thing tomorrow morning. Night!

vsoch:
ok I'm definitely going to need some help to fix the test - it's not intuitive to me how to fake a form submission with files.  Apologies for the trouble!

vsoch:
Woo I fixed it! Hang on...

vsoch:
:green_apple: :green_book: :green_heart: :tea: !!!

chrisfilo:
Looks great, but "upload folder" and "upload .zip" paths gained two clicks. Before it was:
1. Click upload folder
2. Select folder and click upload (in the OS dialog)
3. upload begins

Now we have:
1. Click edit, select upload folder
2. Click choose folder
3. Select folder and click upload (in the OS dialog)
4. Click upload (in the html dialog)
5. upload begins

Could you think of any way of reducing the number of clicks? For example somehow getting rid of step 4. at least for folder and .zip uploads?

vsoch:
Sure, I can think about it. For the old way, finishing with the dialog triggered a form submission. The only way I was able to get the spinner to work was to prevent the default functionality and have finer control of the upload, because between step 3 and 4, there is no way to get feedback from the OS dialog. That's why the spinner couldn't be stopped, and the modal was added (with the extra clicks) to simultaneously control the upload and spinner.

I'm not terribly great with Javascript, and that gives me some confidence that there is very likely an easier solution to all this, I just haven't been able to figure it out.

vsoch:
Question: why was the complicated request function there to begin with? Why can't we just use django native file upload [via forms](https://github.com/axelpale/minimal-django-file-upload-example/blob/master/src/for_django_1-7/myproject/myproject/myapp/templates/myapp/list.html)?

My thinking is that there are two directions to go about doing this. To start from scratch and try to simplify, or take the current and try to hack it to make fewer clicks. My gut is saying that it's getting too complicated.

chrisfilo:
Which function are you referring to?

vsoch:
The [XMLHttpRequest](https://github.com/vsoch/NeuroVault/blob/c45cd2ceba575fcfcbba69b4c18d48cca4f3b540/neurovault/apps/statmaps/templates/statmaps/collection_details.html#L81)

vsoch:
Maybe we can squish it :)

chrisfilo:
I'm not completely sure. @infocortex added it. I think it's there so you could upload a file without displaying a form (which is the default django form mechanism).

vsoch:
ok.

I'm not totally sure what I'm going to do, but I want to start from scratch and gut out most of the page to see if I can simplify things. We can have a hidden form with the data that is submit by simply selecting the form element (call it form) and doing form.submit(). The trick is getting the communication with the spinner. We will figure it out :O)

PR #315, opened 2015-08-09
Update script to read in CognitiveAtlas tasks and contrasts with API

vsoch:
This PR changes the Cognitive Atlas import script to obtain tasks and contrasts programatically, using the API directly instead of an exported json. I anticipate this functionality to be very important in the coming weeks as we update many tasks and contrasts within the Cognitive Atlas. While in the future it would be ideal to query the API directly via the site pages, given that the API is still in beta I believe it safer to store tasks and contrasts in the database. Additionally, the contrast field is still free response (and should be until the user can add a contrast and immediately see it live) and I am leaving the original json of the tasks, just in case.

The API (seems to be) up to date with the [current CognitiveAtlas](http://www.cognitiveatlas.org/tasks/) content:

      >>> CognitiveAtlasContrast.objects.all().count()
      1458  
      >>> from neurovault.apps.statmaps.models import CognitiveAtlasTask
      >>> CognitiveAtlasTask.objects.all().count()
      635

### DIFF
--- a/neurovault/apps/statmaps/migrations/0026_populate_cogatlas.py
+++ b/neurovault/apps/statmaps/migrations/0026_populate_cogatlas.py
@@ -1,22 +1,17 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
-
+from neurovault.apps.statmaps.tasks import repopulate_cognitive_atlas
 from django.db import models, migrations
-import json, os
+import os
 dir = os.path.abspath(os.path.dirname(__file__))
 
+
+# COGNITIVE ATLAS
+###########################################################################
 def populate_cogatlas(apps, schema_editor):
     CognitiveAtlasTask = apps.get_model("statmaps", "CognitiveAtlasTask")
     CognitiveAtlasContrast = apps.get_model("statmaps", "CognitiveAtlasContrast")
-    json_content = open(os.path.join(dir, "cognitiveatlas_tasks.json")).read()
-    json_content = json_content.decode("utf-8").replace('\t', '')
-    data = json.loads(json_content)
-    for item in data:
-        task = CognitiveAtlasTask(name=item["name"], cog_atlas_id=item["id"])
-        task.save()
-        for contrast in item["contrasts"]:
-            contrast = CognitiveAtlasContrast(name=contrast["conname"], cog_atlas_id=contrast["conid"], task=task)
-            contrast.save()
+    repopulate_cognitive_atlas(CognitiveAtlasTask,CognitiveAtlasContrast)
 
 class Migration(migrations.Migration):
 
--- a/neurovault/apps/statmaps/tasks.py
+++ b/neurovault/apps/statmaps/tasks.py
@@ -189,6 +189,36 @@ def save_voxelwise_pearson_similarity_resample(pk1, pk2,resample_dim=[4,4,4]):
             raise Exception("You are trying to compare an image with itself!")
 
 
+# COGNITIVE ATLAS
+###########################################################################
+def repopulate_cognitive_atlas(CognitiveAtlasTask=None,CognitiveAtlasContrast=None):
+    if CognitiveAtlasTask==None:
+        from neurovault.apps.statmaps.models import CognitiveAtlasTask
+    if CognitiveAtlasContrast==None:
+        from neurovault.apps.statmaps.models import CognitiveAtlasContrast
+    
+    import json, os
+    from cognitiveatlas.api import get_task
+    tasks = get_task()
+
+    # Update tasks
+    for t in range(0,len(tasks.json)):
+        task = tasks.json[t]
+        print "%s of %s" %(t,len(tasks.json)) 
+        if tasks.json[t]["name"]:
+            task, _ = CognitiveAtlasTask.objects.update_or_create(cog_atlas_id=task["id"],defaults={"name":task["name"]})
+            task.save()
+            if tasks.json[t]["id"]:
+                task_details = get_task(id=tasks.json[t]["id"])
+                if task_details.json[0]["contrasts"]:
+                    print "Found %s contrasts!" %(len(task_details.json[0]["contrasts"]))
+                    for contrast in task_details.json[0]["contrasts"]:
+                        contrast, _ = CognitiveAtlasContrast.objects.update_or_create(cog_atlas_id=contrast["id"], 
+                                                                                      defaults={"name":contrast["contrast_text"],
+                                                                                      "task":task})
+                        contrast.save() 
+
+
 
 # HELPER FUNCTIONS ####################################################################################
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -47,3 +47,4 @@ git+https://github.com/trungdong/prov.git#egg=prov
 opbeat
 djrill
 django-hstore==1.3.5
+cognitiveatlas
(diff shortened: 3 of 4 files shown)


chrisfilo:
Great to see this coming together! Could you turn it into a function that
would be called by the script as well as this migration
https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/migrations/0026_populate_cogatlas.py?
This way new installs will be always up to date.

vsoch:
Good idea.

vsoch:
Looks like the migration worked ok with the function, but will be good to get others to test because I was having URL errors on my computer! It likely is because I have web-scraping going on, but I want to be absolutely sure it's not something about the Cognitive Atlas API.

chrisfilo:
I'm getting the following error:	
	
	$ docker-compose run --rm uwsgi bash -c "./manage.py shell < scripts/repopulate_cognitive_atlas_tasks.py"
	Warning: the mapping "create_hstore_extension.sh:/docker-entrypoint-initdb.d/create_hstore_extension.sh:ro" in the volumes config for service "db" is ambiguous. In a future version of Docker, it will designate a "named" volume (see https://github.com/docker/docker/pull/14242). To prevent unexpected behaviour, change it to "./create_hstore_extension.sh:/docker-entrypoint-initdb.d/create_hstore_extension.sh:ro"
	Warning: the mapping "nginx.conf:/etc/nginx/conf.d/default.conf:ro" in the volumes config for service "nginx" is ambiguous. In a future version of Docker, it will designate a "named" volume (see https://github.com/docker/docker/pull/14242). To prevent unexpected behaviour, change it to "./nginx.conf:/etc/nginx/conf.d/default.conf:ro"
	Warning: the mapping "uwsgi_params.par:/etc/nginx/uwsgi_params.par:ro" in the volumes config for service "nginx" is ambiguous. In a future version of Docker, it will designate a "named" volume (see https://github.com/docker/docker/pull/14242). To prevent unexpected behaviour, change it to "./uwsgi_params.par:/etc/nginx/uwsgi_params.par:ro"
	/usr/local/Cellar/fig/1.4.0/libexec/vendor/lib/python2.7/site-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
	  InsecurePlatformWarning
	Python 2.7.10 (default, Jul 18 2015, 00:45:22)
	[GCC 4.9.2] on linux2
	Type "help", "copyright", "credits" or "license" for more information.
	(InteractiveConsole)
	>>> >>> http://cognitiveatlas.org/api/v-alpha/task?
	Result Includes:<pandas:data frame><json:dict><txt:str><url:str>
	0 of 635
	1 of 635
	http://cognitiveatlas.org/api/v-alpha/task?id=trm_523c7a0a73cf5
	Result Includes:<pandas:data frame><json:dict><txt:str><url:str>
	Found 5 contrasts!
	  File "<console>", line 1, in <module>
	  File "/code/neurovault/apps/statmaps/tasks.py", line 218, in repopulate_cognitive_atlas
	    "task":task})
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/manager.py", line 92, in manager_method
	    return getattr(self.get_queryset(), name)(*args, **kwargs)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 439, in update_or_create
	    obj, created = self._create_object_from_params(lookup, params)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 457, in _create_object_from_params
	    obj.save(force_insert=True, using=self.db)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 589, in save
	    force_update=force_update, update_fields=update_fields)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 617, in save_base
	    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 698, in _save_table
	    result = self._do_insert(cls._base_manager, using, fields, update_pk, raw)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/base.py", line 731, in _do_insert
	    using=using, raw=raw)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/manager.py", line 92, in manager_method
	    return getattr(self.get_queryset(), name)(*args, **kwargs)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/query.py", line 921, in _insert
	    return query.get_compiler(using=using).execute_sql(return_id)
	  File "/usr/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 921, in execute_sql
	    cursor.execute(sql, params)
	  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 81, in execute
	    return super(CursorDebugWrapper, self).execute(sql, params)
	  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
	    return self.cursor.execute(sql, params)
	  File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__
	    six.reraise(dj_exc_type, dj_exc_value, traceback)
	  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
	    return self.cursor.execute(sql, params)
	DataError: invalid input syntax for integer: "trm_523c7a0a73cf5"
	LINE 1: ...S ('Irritability/Agitation', 'cnt_523c7cdacdc20', 'trm_523c7...
	                                                             ^

Maybe it's due to recent updates in Cognitive Atlas or its API?

vsoch:
I'm still a few hours from home but I can look when I get there! I saw a similar error to this when there was a break in my internet, and for some reason it corrupted the function to add to the database (because when I tested a simple manual string it would trigger an error). I was not able to reproduce it when I had s good connection (see there is something about urllib3). The functionality to do the actual update or create is totally the same, and the only difference is obtaining the data via the internet.

vsoch:
Could there be some kind of conflict between the function cogat is using to get the result and some version installed in NV?

vsoch:
I have been having trouble with urllib3 and my CogatPheno, which is what makes me think of version conflicts.

chrisfilo:
The urllib3 warning is due to recent docker update - probably unrelated.

I got this error on a freshly built container.

chrisfilo:
I think it's because the initial migrations did not work (due to network issues). Let me try again.

vsoch:
Could it have to do with timing then - if the internet connection is slow and the call to the database happens before the result is completely ready, it does this? If you still get it, maybe it would be worth trying slowing down the process a bit with a pause?

vsoch:
We could try doing all the queries first, saving, then inputting into the database.

vsoch:
Regardless if it works on the circle server, it's an issue if it breaks in development. I'm having trouble consistently reproducing the error.  I think the best way to debug this is to rewrite the function to do the queries separately from the updating. I'm also not totally trusting of the update function because it had unexpected result last week when I was trying it with CogatPheno (creating new entries entirely when it should have been updating) so I would also re-write it to just do the update manually. So for now,  let's just let this PR sit, I'll give it another try when I can :O)

PR #330, opened 2015-08-31
Update/collection view

vsoch:
This draft PR will update the collection view to dynamically view images:

empty view
empty

with images
view

I have also refactored the entire page to be html instead of haml.

The papaya API is limited to adding and substracting images, so for now multiple images can be added but I can't figure out how to manipulate their properties (eg transparency) so we can see all the layers.

The PR isn't ready for merge just yet - I need help with the following:
- the datatables pagination isn't working (likely something to do with a style or js file, ordering) and I can't figure it out. I figure someone's else's eyes on it would do good. 
- It might just be my dev environment, but it seems like images from old collections aren't being deleted, and they get added to new collections. Can you test this on your end?

Everything else other than that is working (there were a lot of bugs figuring this out!)


vsoch:
This change would make the image viewer on the single image pages (currently still there) redundant, and my thinking is that this "single image" view page would be replaced with more of a "work with" annotation / metadata page.

chrisfilo:
I would keep the single image page for now - it's essential for a few things:
- sending links to a particular image
- editing a single image
- decoding an image
- viewing metadata of an image

When you say "images from old collections aren't being delete" what did you mean by "old collections"? Or what are the steps to replicate this?

vsoch:
Sweet, the fix worked! I submit the change to the Dockerfile (adding the libgeos-dev) as a separate PR. 

To replicate the weird behavior:
- make a collection
- add images to it
- make another collection
- add images too it
- there are more images added than were in the folder

chrisfilo:
So the images from the first collection show up in the second collection?

chrisfilo:
I tried to replicate, but could not. This is what I did:
1. Added a collection test1
2. Uploaded a folder with 6 images to test1
3. Added a collection test2
4. Uploaded a folder with 6 images

Both collection have 6 images and nothing is doubled.

vsoch:
ok great! It's just me then. I'll be able to do the detail work for this PR likely next week.

vsoch:
Ah just fixed the datatables! I didn't initialize them, duh. /bonk

vsoch:
ok datatables are fixed (pagination is working) as are some of the bugs for the papaya viewer. It still doesn't display transparency - could it be that they are added as images and not overlays?

chrisfilo:
Yes I think you need to add the anatomical image each time you reset the viewer.

vsoch:
It does include the anatomical - if you look at [the function here](https://github.com/vsoch/NeuroVault/blob/update/collection-view/neurovault/apps/statmaps/templates/statmaps/collection_details.html#L56)

we grab the images from the parameters (including the MNI)

       // Get current images
      var papayaimages = params["images"];

We then get the currently selected image      

      var contenderimage = $(mrimage).attr("filename");
      
Then it's easiest to just remove this image from the current list, if it's there

      // Remove selection
      papayaimages = jQuery.grep(papayaimages, function(value) {
          return value != contenderimage;
      });

Then we make sure buttons correspond with the selection (ie, if a button was pressed "active" we just remove the active label so it doesn't look pressed. If the button was not pressed, we add the active class and add the image to the list):

      if ($(mrimage).hasClass("active")){
          $(mrimage).removeClass("active")        
      } else {
          $(mrimage).addClass("active")

And this is the line that pushes the new image to the array (and note that we never remove the original anatomical image)

          papayaimages.push(contenderimage);
      }

Now we reset the viewer entirely, and each image in the array (including the anatomical) is re-added:

     // Reset viewer and load new images
     papayaContainers[0].viewer.resetViewer();
     papayaContainers[0].toolbar.updateImageButtons();
     for (i = 0; i < papayaimages.length; i++) { 
        console.log("Adding image " + papayaimages[i]);
        papayaContainers[0].viewer.loadImage([papayaimages[i]], true)
     }
   }

If the "loadImage" function is not equivalent to setting the images in the params, that may be the issue. The papaya API doesn't have any functions for interactively changing the image data, so it's pretty limited.

chrisfilo:
Hmm... when I tested it (a few commits ago) it behave as if everything was removed and just a single statistical map was added when you press "View". When there are more than one images in papaya there are multiple colormap buttons on the top toolbar - there was only one when I tested it. Maybe there is something wrong with the way you get currently loaded images?

vsoch:
Yeah that must be it.

vsoch:
ok - this fix will make the viewer display the currently selected image on the standard anatomical - basically what we have now. We can have this version until I figure out how to nicely display multiple images. I'm thinking it's just a matter of adding several overlays, but need to do more testing.

vsoch:
[functions I found](https://github.com/NeuroVault/NeuroVault/commit/ff3f464e9c495343f1e9386f725b11a8b3e15172#diff-1c26b88e70f727d2d1fd56d3c64a3993R62)

chrisfilo:
I tested the latest version and run into three issues:

1. the file motor_lips_smoothed is displayed in a funny way on the collection view even though it looks good on a single map view (download: https://docs.google.com/uc?id=0B77zr9yIiKOTWWo5cU1veno0d28&export=download)
screen shot 2015-09-07 at 15 55 44

2. collection view uses split colormaps instead of "combineParametric" (like the single map view - see https://github.com/NeuroVault/NeuroVault/blob/4d6bcde4b05a69a9f3c78747e4f7bc59ac7f3034/neurovault/apps/statmaps/templates/statmaps/statisticmap_details.html.haml#L12)

3. Papaya menus are positioned outside of the viewer:
screen shot 2015-09-07 at 15 51 49

I also noticed a strange behaviour where viewing maps stops working (you click "view" and only anatomical is being shown). I happened after clicking around a bit, but I could not find a reliable pattern to replicate this (there is also no error in the JS console). Have you experienced this?

I would also suggest to load the first map by default when the collection view loads for the first time. Otherwise excellent work!

vsoch:
thanks :O)

vsoch:
Ok, the bugs should be resolved! Here are some details:

* motor_lips_smoothed was displaying strange as a result of missing "symmetric":true as one of the parameters for the image. This value:

  >> sets the negative range of a parametric pair to the same size as the positive range.

and this means that, for that particular smoothed image, one of the ranges was likely much different than the other, and making the entire image appear gray.

* Papaya viewer controls: This bug was bad. I tried everything from customizing the css, to using Jquery to do it, and while it worked when I copy pasted "click" functions into my local browser, no matter where I put them in the script papaya would seem to "redo" it's (broken) styling. After two hours of this nonsense, and realizing that the style was equivalent to the single image viewer (that did not have buggy menus) I realized that it must have something to do with its placement in the page. I found that, when I removed the viewer from the col-md-6, the style was perfect, and I couldn't tell you why - even "col-md-12" (a full page width) was broken, but placing the viewer in a row sans column worked fine! So the final fix was to "elicit" something that looked like a column by using an aside class instead. 

* Strange behavior: I was not able to reproduce it, but I hope it was a one time thing! :)

chrisfilo:
Cool - works as advertised! Did you try to make first image in the collection load into papya on start? Currently when the page loads for the first time the viewer is empty. 

We can also use the eye  and eye-slash icons instead of the view button from fontawesome: http://fortawesome.github.io/Font-Awesome/icon/eye/

vsoch:
oh, heh, I actually just forgot to do that :)

I'm working on the second PR now, but those fixes seem pretty reasonable, and I'll finish up the second and go back to them!

vsoch:
Font awesome!

image

and this is the view on page load.

vsoch:
How is this logical? I didn't change anything but javascript!

vsoch:
It looks like we ran out of space!

      OSError: [Errno 28] No space left on device: '/tmp/tmpne7zFZpip-wheel-'

I'm retrying without cache, if that makes a difference. This is nuts!

vsoch:
Ah, maybe it is because circle only gives one container. Then running two builds at the same time is not a good idea :)

PR #335, opened 2015-09-10
Update/cognitive atlas contrast

vsoch:
This PR will add contrast ids to be saved with the images. This workflow is as follows:

On edit image view, user sees option to select a task:

image

User selects a task. Case 1: the task does not have contrasts defined in Cognitive Atlas:

image

Nothing changes - the user can still write in a contrast definition.

Case 2: user selects a task with contrasts in the cognitive atlas:

image

The "write in" box disappears, and a new dropdown appears that is populated with the contrasts, each name is associated with the contrast ID and the new field "cognitive_contrast_cogatlas" in the database! If the user then browses and the contrast is not there, there is an "Other" option:

screenshot from 2015-09-09 17 05 57

Selecting "Other" reveals again the write in, and the write in is hidden if other is deselected:

image

### Details

- A new field, "[cognitive_contrast_cogatlas](https://github.com/vsoch/NeuroVault/blob/update/cognitive-atlas-contrast/neurovault/apps/statmaps/models.py#L515)" was added to the statistical maps to hold a CognitiveAtlasContrast. I chose to add a new field and maintain the write in "contrast_definition" so that the user would have the option to select a defined contrast or write in a custom one. This risks having redundant information, but:
    - We can assume that if a cognitive_contrast_cogatlas is defined, it takes preference over the write in
    - For our current images, I will manually go through and define these fields (to overwrite the write ins)
    - the interface will only reveal the write in box for an image with task with contrasts defined in the cognitive atlas given that "other" is selected for the "cognitive_contrast_cogatlas" variable, so it will not be possible for a contrast with unique ID to exist with a write in.
    - The API [now reveal both](https://github.com/vsoch/NeuroVault/blob/update/cognitive-atlas-contrast/neurovault/urls.py#L149) for this transition, so worst case scenario both pieces of information are available to the user. When NeuroVault is better integrated with a real time Cognitive Atlas API, we can completely remove the write in, and instead have the user add the new contrast. 
  - Given that we are now using contrasts, I think there should be some kind of cron job on the server to minimally update the contrasts/tasks each night.

This PR also includes the previous to edit the collection details, without showing the first image. I will address that in the other PR.


chrisfilo:
I just tested it by trying to add one image (collections/1/addimage) and the "Cognitive Atlas Contrast" field is visible all the time listing all contrasts from all tasks. I guess that's a bug?

vsoch:
The field itself should always be visible. Could you give more detail to the circumstances when it shouldn't be?

vsoch:
Oh I see, I didn't realize that the user could also specify a contrast on the "add one image" page (the work I did is entirely on the "edit image" page.) I will fix this promptly and update the PR - is there anywhere else with contrast specification?

vsoch:
ok, the contrast selection has been added to the single image view, as well as the "add image for neurosynth" but I'm not sure where this page is to review it. Since we are using code over, I chose to move the functionality into another template, `_contrast_lookup.html` that can be included in both pages.

chrisfilo:
Ok I tried it again (collections/1/addimage). The "Cognitive Atlas Contrast" field does not appear no matter what task I select. There are no error messages in the javascript console. This test was done after a fresh restart of docker-compose.

vsoch:
The field you should see is "Cognitive Paradigm" and only when you select one that has contrasts does it show up.

Also make sure that you have run the migration that updates contrasts, and adds the contrast field.  There is clearly something that is missing on your end because it is working for me...

chrisfilo:
I removed all the containers (docker-compose rm) and recreated them (docker-compose up). All migrations have been applied automatically and without errors (including 0052_statisticmap_cognitive_contrast_cogatlas). I was trying several tasks including "Stop-signal task" which does have contrasts defined in cognitive atlas.

Did you add all files to your commits? Can you try removing your containers and starting afresh?

vsoch:
Yes I can do that - I'm sorry it's not working we will figure out what's going on... it is a very important PR! :o)

I still have about 45 minutes to walk home but will do this promptly tonight.

vsoch:
Ah... Going to be a lot longer than that... I left my dinner in the LKSC. Oups. :)

vsoch:
ok, I just reproduced the error. I think the issue is that the migration to link the contrast to the task isn't done until after the migration that updates task/contrasts, so when the tasks are defined they don't have links to any contrasts (try clicking "view source" on the page, and look for a variable called 'contrast_lookup' - you will see something like this:

      var contrast_lookup = {"trm_553fc63a54ae6": [], "trm_5346927710e88": [], "trm_4f241fd47f16b": [], "trm_4c8a830dec136": [], "trm_5106eee90937a": [], "trm_524b4a402c87e": [], "trm_4da86cdbd9eeb": [], "trm_551f0c294ce23": [],

all of those empty [] should have the contrasts inside, and had the migration to add the contrast to the task been run before the migration to update the tasks, those would be populated.  

So the fix is to run the "update cognitive atlas" migration after the field has been added to the task. How do we alter the order of the migrations?

vsoch:
And minimally, if you want to review the PR, just run the script manually to update tasks :)

chrisfilo:
We cannot change the order since those migrations have already been deployed on the server. The update should be done by just adding new migrations.

I don't quite understand why the order matters in this case. Are you sure this is the problem?

vsoch:
Yes I'm sure, because I just ran the script manually:

      docker-compose uwsgi run bash
      python manage.py shell < scripts/repopulate_cognitive_atlas.py

and it is now working - the concept_lookup is populated:

image

vsoch:
I agree that it is not logical.

vsoch:
ok, I think I have a fix. It turns out, migrations aren't linked to the database in any way - it's a record of the application kept in the migration scripts, and the order isn't even important with regard to the numbers - the only thing that matters is the dependencies. This means that we can make our own script files, and Django even makes it easy to make a "blank" migration:

      python manage.py makemigrations --empty yourappname

What I did is to add a migration that runs the repopulate_cognitive_atlas.py after the edit to the database. I'm still not sure why this was an issue, but it resolved the problem so it seems to be a reasonable solution. I think I even anticipated needing to do this from the migration, because I gave the tasks.py script [an option to take in the CognitiveAtlasTask and CognitiveAtlasContrast objects as variables](https://github.com/vsoch/NeuroVault/blob/master/neurovault/apps/statmaps/tasks.py#L194).

This resolves the issue in terms of starting from scratch and NeuroVault loading with everything working. What I'm worried about is if the cognitiveatlas API calls are going to work on circle (last time if I remember correctly they were not, which is why we stuck to reading from the json for the initial load. That would be quite the catch-22. If it doesn't work on circle, our only option may be to figure out a way to ensure a "manual" run of updating the tasks and contrasts (at the end) after the initial install.

chrisfilo:
I would prefer the migrations not to depend on cognitive atlas api (its
slow and it requires internet connectivity which limits the development
process). I hope there is another way.

We should try to understand why repopulating cognitive atlas is necessary
to begin with.

vsoch:
I don't know. I've done my best on this. I had wanted this analysis to be completely programmatic but it doesn't look like that is going to happen anytime soon. I'm switching to plan B.

chrisfilo:
I've sent you a pull request with a fix to this problem: https://github.com/vsoch/NeuroVault/pull/9

I also noticed one more thing. When I edit an image (after adding it) the cognitive atlas contrast drop list does not appear until I change the task even though I set the contrast before (and the task had contrasts defined). Hopefully this will be easy to fix.

chrisfilo:
I might also make sense to add the "Cognitive Atlas Contrast" column to the tabular interface @burnash contributed.

vsoch:
ok, I've modified the PR so that a selected contrast will show up first. 

It's not a simple issue to just add the column to burnash interface. Given a collection with images spanning tasks, this would mean having custom select menus for each row in the table, and then needing to remove the select for specific rows for which the task does not have contrasts defined. Actually, we wouldn't want that field there at all - there would be no values to populate it. We can't open it up to include any contrast name because then a task could be tagged with a contrast that does not belong to it.

From a usability standpoint, I don't see a user wanting to manually select contrasts in a batch fashion. If anything, that level of specificity would just be annoying if I want to do a batch data input. It almost seems more reasonable to remove contrast specification entirely from the batch, since it might even just get in the way. For now, it seems harmless to have the write in field (contrast_definition), and I think we should leave as is.

chrisfilo:
Thanks!

re: tabular interface - I see now that this would require quite some development. The only situation where this would be useful is for single subject maps.

PR #336, opened 2015-09-10
moving buttons to the right side to balance the UI

vsoch:
Here is a PR if you want to move buttons to the right side.


vsoch:
For documentation - here is the current:

image

and the proposed update:

withauthorsloggedin
withauthorsnotloggedin

I really like the :eyes: that was a good idea.

PR #339, opened 2015-09-15
Make the main page order collection by when the DOI was added

chrisfilo:


chrisfilo:
@vsoch could you have a look?

vsoch:
yep

vsoch:
Ok I tested:

- Adding as public without DOI: verified not in list
- Adding as private without DOI: verified not in list
- Added DOI to private: verified still not in list
- Made public the private with DOI: verified in list
- Adding as public with DOI: verified in list
- Adding as public without DOI: verified not in list

And the order seems appropriate - newer at the top. My one concern was for our current collections that DO have DOI, but are private - the doi is already there so the doi date should be set to when the collection was made, but when the collection goes public it will be set to the time it goes public. If that is what we want, looks good to me! If not, maybe we should set DOI date for the currently private collections to the collection creation date (so remove the check for private in the migration).

vsoch:
but thinking about this - it depends if you want the DOI date to be actually when it was added (what I was thinking above) OR when it is "publicly added" in which case the way it is now would be desired.

vsoch:
I think we want the latter, the DOI date when it is publicly added.

chrisfilo:
Thanks!

PR #340, opened 2015-09-17
Fix/papaya viewer

vsoch:
This will fix the mother :truck: er papaya viewer! The viewer .js has been updated to the current version released today, with a new "resetViewer" function. We can either reset the viewer to use old parameters, or specify new ones. I tried still initializing the viewer with the image parameters (including the color luts) and then reloading, and this is when we see the error of "missing" a lut. I also tried folding the params into a function, and couldn't get it working quite right. When I [redefine the luts on the fly](https://github.com/vsoch/NeuroVault/blob/fix/papaya-viewer/neurovault/apps/statmaps/templates/statmaps/collection_details.html#L58) (fresh params!) and then reset the viewer:

            params = []
            ... redefine here ...
            papaya.Container.resetViewer(0, params);  // specify new ones

Then the bug seems to go away! This should work.


chrisfilo:
This did the trick. Excellent work!

PR #357, opened 2015-10-24
Support for object level permission

chrisfilo:
@burnash This implements object level permissions which should make writing the API easier. Please let me know if this works for you.


burnash:
@chrisfilo If I get it right this PR delegates all permission checks for owners and contributors to Guardian, but reuses the existing UI. Very good. Works for me.

chrisfilo:
One thing I could not figure out is the "add" permission on object level.
I'm not sure if this will be a problem with the API. We shall see!

PR #362, opened 2015-10-27
WIP: Image upload API

burnash:
@chrisfilo This implements the image upload API. Although this is a second part of the "Auth & Image Upload" feature I'd like to put it on review first while I finish tests and a summary for [OAuth Tookit integration](https://github.com/burnash/NeuroVault/tree/d9ae8d6ae548820919acfba29c31850b531c53c1).

### DIFF
--- a/neurovault/urls.py
+++ b/neurovault/urls.py
@@ -19,7 +19,7 @@ from rest_framework.renderers import JSONRenderer
 from django.http import Http404, HttpResponse
 from rest_framework.response import Response
 from rest_framework.views import APIView
-from rest_framework import permissions
+from rest_framework import permissions, status
 from oauth2_provider import views as oauth_views
 import xml.etree.ElementTree as ET
 from taggit.models import Tag
@@ -148,6 +148,11 @@ class ImageSerializer(serializers.HyperlinkedModelSerializer):
             if pd.isnull(val):
                 orderedDict[key] = None;
         return orderedDict
+
+
+class EditableStatisticMapSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = StatisticMap
 
 
 class StatisticMapSerializer(ImageSerializer):
@@ -408,6 +413,18 @@ class CollectionViewSet(mixins.RetrieveModelMixin,
         serializer = ImageSerializer(
             page, context={'request': request}, many=True)
         return paginator.get_paginated_response(serializer.data)
+
+    @detail_route(methods=['post'], url_path='addimage')
+    def add_image(self, request, pk=None):
+        collection = get_collection(pk, request, mode='api')
+        image = StatisticMap(collection=collection)
+        serializer = EditableStatisticMapSerializer(data=request.data,
+                                                    instance=image)
+        if serializer.is_valid():
+            serializer.save()
+            return Response(serializer.data,
+                            status=status.HTTP_201_CREATED)
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
     def retrieve(self, request, pk=None):
         collection = get_collection(pk, request, mode='api')
--- a/neurovault/urls.py
+++ b/neurovault/urls.py
@@ -13,6 +13,7 @@ from lxml.etree import xmlfile
 admin.autodiscover()
 from rest_framework import viewsets, routers, serializers, mixins, generics
 from neurovault.apps.statmaps.views import get_image, get_collection
+from neurovault.apps.statmaps.views import owner_or_contrib
 from rest_framework.decorators import detail_route, list_route
 from django.contrib.auth.models import User, Group
 from rest_framework.renderers import JSONRenderer
@@ -153,6 +154,7 @@ class ImageSerializer(serializers.HyperlinkedModelSerializer):
 class EditableStatisticMapSerializer(serializers.ModelSerializer):
     class Meta:
         model = StatisticMap
+        read_only_fields = ('collection',)
 
 
 class StatisticMapSerializer(ImageSerializer):
@@ -417,6 +419,10 @@ class CollectionViewSet(mixins.RetrieveModelMixin,
     @detail_route(methods=['post'], url_path='addimage')
     def add_image(self, request, pk=None):
         collection = get_collection(pk, request, mode='api')
+
+        if not owner_or_contrib(request, collection):
+            self.permission_denied(request)
+
         image = StatisticMap(collection=collection)
         serializer = EditableStatisticMapSerializer(data=request.data,
                                                     instance=image)
--- a/neurovault/urls.py
+++ b/neurovault/urls.py
@@ -1,35 +1,35 @@
-from neurovault.apps.statmaps.voxel_query_functions import voxelToRegion, getSynonyms, toAtlas, getAtlasVoxels
+from neurovault.apps.statmaps.voxel_query_functions import (
+    voxelToRegion, getSynonyms, toAtlas, getAtlasVoxels
+)
 from rest_framework.relations import StringRelatedField, PrimaryKeyRelatedField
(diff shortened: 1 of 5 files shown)


neurovault/urls.py
```diff
@@ -380,6 +389,7 @@ class CollectionViewSet(mixins.RetrieveModelMixin,
     filter_fields = ('name', 'DOI', 'owner')
     serializer_class = CollectionSerializer
     filter_backends = (DjangoFilterBackend,)
+    permission_classes = (permissions.IsAuthenticatedOrReadOnly,)
```
chrisfilo: shouldn't this be DjangoObjectPermissions? (same for collections)

burnash: Yeah, I tried that. Problem is, when we're handling image upload in `CollectionViewSet`, DRF first checks permissions for the `CollectionViewSet` while [dispatching a request](https://github.com/tomchristie/django-rest-framework/blob/4e5da16961c8f11d01a2f9dc475efacb29e778c7/rest_framework/views.py#L454) and before calling the handler. So when it hits `APIView.check_permissions()`, it calls [`has_permission`](https://github.com/tomchristie/django-rest-framework/blob/4e5da16961c8f11d01a2f9dc475efacb29e778c7/rest_framework/views.py#L318) (and not [`has_object_permission`](https://github.com/tomchristie/django-rest-framework/blob/4e5da16961c8f11d01a2f9dc475efacb29e778c7/rest_framework/views.py#L329)). `DjangoObjectPermissions` doesn't implement `has_permission()` but since it's derived from `DjangoModelPermissions`, `DjangoModelPermissions.has_permissions()` gets called. And it's irrelevant to the job.

[As DRF's documentation says](http://www.django-rest-framework.org/api-guide/permissions/#custom-permissions): "The instance-level `has_object_permission` method will only be called if the view-level `has_permission` checks have already passed." The `permission_classes` in question is a view-level attribute.

Still even if we disable the irrelevant view-level checks, I can't see any way to trigger `has_object_permission` in the `detail_route` other than calling [`GenericAPIView.get_object`](https://github.com/tomchristie/django-rest-framework/blob/6284bceaaff0e53349131164ce5c16cda8deb715/rest_framework/generics.py#L76) (instead of `get_collection`.)

So I decided to go with an explicit check and used the same method for exception raising as in `APIView.check_object_permissions `. 

While writing this I realized that there's another option available (for more implicit checks):
1. Set `permission_classes` to `DjangoObjectPermissions`, as you suggest.
2. In the `CollectionViewSet.add_image` endpoint, call `GenericAPIView.get_object()` to fetch the collection object (this will trigger `check_object_permissions`.)
3. Use PUT method instead of POST for images, because it maps to `change_collection` in [`DjangoObjectPermissions.perms_map`](https://github.com/tomchristie/django-rest-framework/blob/6284bceaaff0e53349131164ce5c16cda8deb715/rest_framework/permissions.py#L88).
4. Use Django's model permissions system to pass the view-level `has_permission` check. (I can't find any trace of a model permissions usage in the code. But I may miss something.)

neurovault/urls.py
```diff
@@ -400,6 +410,22 @@ def images(self, request, pk=None):
             page, context={'request': request}, many=True)
         return paginator.get_paginated_response(serializer.data)
 
+    @detail_route(methods=['post'], url_path='addimage')
```
burnash: I also haven't figured out yet how to make two "image" `detail_route`s co-exist. Initially I declared the add image endpoint as `POST /collections/<id>/images` with `@detail_route(methods=['post'], url_path='images')` but it turned out to override the `images()` endpoint (one that serves `GET`) for some unknown reason. I may need to study DRF more to figure out why.

neurovault/urls.py
chrisfilo: Since there is not a clear way to map addition of related object to django object permissions I think your solution is adequate. Hopefully the object level permission will get handy when we start modifying or deleting objects.

burnash: Object level permission are already in handy even though I use them indirectly. As the API expands I expect object level permission's way to scale better than equality checks.

burnash:
@chrisfilo I rebased the branch to the latest master.

neurovault/urls.py
```diff
@@ -409,6 +419,21 @@ def images(self, request, pk=None):
...
 
+    def add_image(self, request, pk=None):
+        collection = get_collection(pk, request, mode='api')
+
+        if not owner_or_contrib(request, collection):
+            self.permission_denied(request)
+
+        image = StatisticMap(collection=collection)
+        serializer = EditableStatisticMapSerializer(data=request.data,
+                                                    instance=image)
```
chrisfilo: This works only for StatisticalMaps. What about Atlases and NIDM-Results? Maybe we should have different endpoints for each class.

burnash: What is a proper way to differentiate between those other classes? Different endpoints is the simplest way to implement it. 
Another simple way is to require an additional argument `image_type` and to do something similar to this lines: https://github.com/NeuroVault/NeuroVault/blob/48743b44f04d37ab1e88ca8cc292ff19dc865be6/neurovault/urls.py#L134

What will be the most convenient way for a future user? How hard it'll be to guess the image type?

chrisfilo: Since NIDM results is not an image_type per se I would prefer going with three different endpoints:

* http://neurovault.org/api/images/ (for statmaps)
* http://neurovault.org/api/atlases/ (for atlases)
* http://neurovault.org/api/nidm_results/ (for NIDM results)

burnash: To clarify, the URL used for posting images in this PR is 

    http://neurovault.org/api/collections/<id>/images/

So it's nested under collection endpoint and gets the collection id from the url itself. Do you want me to change it to

    http://neurovault.org/api/images/

Along with new endpoints, or to make these new endpoints nested under the collection URL:

    http://neurovault.org/api/collections/<id>/atlases/
    http://neurovault.org/api/collections/<id>/nidm_results/

It looks like the latter is faster to implement since it'll require less changes.

chrisfilo: If it's easier to implement I'm happy to go with your suggestion.

burnash:
@chrisfilo Added api endpoints for atlases and NIDM-results.

chrisfilo:
This looks good, but there are a bunch of routines in https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/forms.py#L673 and https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/nidm_results.py that do not seem to be triggered when uploaded through the API. They are meant to extract maps from the .nidm.zip files as well as modify and update the ttl inside the .nidm.zip to fix some known bugs as well as update URLs to the neurovault ones.

Feel free to refactor this code if you think it's a better way to go. NIDM handling has a reasonably good test coverage.

burnash:
@chrisfilo I see. There's indeed a need for some sort of factoring out the validation logic of `NIDMResultsForm`, because I don't see any way to plug the form directly to `NIDMResults` serializer.

burnash:
@chrisfilo I refactored `NIDMResultsForm` and added the additional cleaning to the NIDM upload handler. Please review.

burnash:
@chrisfilo It looks like [`ImageForm.clean`](https://github.com/NeuroVault/NeuroVault/blob/b3a857cf25ab9a67502ed2643336a5a58662c113/neurovault/apps/statmaps/forms.py#L354) has some utility code as well. I'll put that to a serializer. Not sure how to handle `self._errors` in relation to the DRF's serializer.

burnash:
@chrisfilo I plugged in form validation for both statmaps and atlases. It works similar to NIMD upload validation handling. This ready for review as well.

chrisfilo:
Excellent! Works as advertised!

PR #365, opened 2015-10-29
WIP: API authorization and collections creation

burnash:
This adds authorization to the API with help of [Django OAuth Toolkit](https://github.com/evonove/django-oauth-toolkit). The PR also implements collection creation with `POST` to `/api/collections/`.

Since CORS in NeuroVault implemented on the Nginx level, the settings differ from the ones in the Toolkit's [tutorial](https://django-oauth-toolkit.readthedocs.org/en/latest/tutorial/tutorial_01.html).

PR #366, opened 2015-10-30
User settings pages with tabs

burnash:
This PR replaces current "edit profile" page with two "settings" pages. Settings are split into tabs. The interface for managing OAuth applications from #365 will go into another tab in settings.

The PR renames a link to edit user profile page to "Settings":
screen shot 2015-10-31 at 00 39 55

I left the `edit_user.html` intact because I already have it restyled in another branch (sign-in & sign-up.)

### DIFF
--- a/neurovault/apps/users/forms.py
+++ b/neurovault/apps/users/forms.py
@@ -1,7 +1,8 @@
 from django import forms
 from django.contrib.auth.models import User
-from django.contrib.auth.forms import UserCreationForm, UserChangeForm,\
-    PasswordChangeForm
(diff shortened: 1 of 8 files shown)


neurovault/apps/users/forms.py
```diff
@@ -16,20 +17,35 @@ def save(self, commit=True):
         if commit:
             user.save()
         return user
-    
-class UserEditForm(UserChangeForm):
```
burnash: @chrisfilo I have to change inheritance to plain ModelForm, because I see no other way to discard the read-only password field.

PR #388, opened 2015-12-16
WIP: Update and delete methods for collection and images API

burnash:
This adds update, partial update and delete methods (HTTP put, patch & delete) to the collection API and collection items API.

### DIFF
--- a/neurovault/api/permissions.py
+++ b/neurovault/api/permissions.py
@@ -1,7 +1,35 @@
-from rest_framework.permissions import DjangoObjectPermissions
+from rest_framework.permissions import (DjangoObjectPermissions,
+                                        SAFE_METHODS, Http404)
+from guardian.ctypes import get_ctype_from_polymorphic
 
 
 class ObjectOnlyPermissions(DjangoObjectPermissions):
 
     def has_permission(self, request, view):
+        return True
+
+
+class ObjectOnlyPolymorphicPermissions(ObjectOnlyPermissions):
+
+    def has_permission(self, request, view):
+        return True
+
+    def has_object_permission(self, request, view, obj):
+        ctype = get_ctype_from_polymorphic(obj)
+
+        model_cls = ctype.model_class()
+        user = request.user
+
+        perms = self.get_required_object_permissions(request.method, model_cls)
+
+        if not user.has_perms(perms, obj):
+            if request.method in SAFE_METHODS:
+                raise Http404
+
+            read_perms = self.get_required_object_permissions('GET', model_cls)
+            if not user.has_perms(read_perms, obj):
+                raise Http404
+
+            return False
+
         return True
--- a/neurovault/api/permissions.py
+++ b/neurovault/api/permissions.py
@@ -1,0 +1,7 @@
+from rest_framework.permissions import DjangoObjectPermissions
+
+
+class ObjectOnlyPermissions(DjangoObjectPermissions):
+
+    def has_permission(self, request, view):
+        return True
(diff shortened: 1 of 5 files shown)


chrisfilo:
Do I understand correctly that this is directly using the object level permissions?

burnash:
@chrisfilo It looks like I missed the object level permissions entirely.

I checked again how to turn the object level permissions on, and although it appears quite simple and straightforward, there is one catch I can recall from [my previous permissions-in-drf encounter](https://github.com/NeuroVault/NeuroVault/pull/362#discussion_r43209433).

The simplest way is to use a [DjangoObjectPermissions](http://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions) class as a permission class in a ViewSet:

```python
class ImageViewSet(...):
    ...
    permission_classes = (permissions.DjangoObjectPermissions,)
```

That should handle object permissions directly. However, `DjangoObjectPermissions` requires Django's standard `django.contrib.auth` model permissions. This is because in DRF permissions are [checked at the initial request stage](https://github.com/tomchristie/django-rest-framework/blob/4e5da16961c8f11d01a2f9dc475efacb29e778c7/rest_framework/views.py#L377) and the `DjangoObjectPermissions` class falls back to `DjangoModelPermissions` for handling that. 

Is NeuroVault using model permissions? I can't find any signs of it in the code. I might be wrong though.

If NeuroVault is not using model permissions, the way to go is to override `DjangoObjectPermissions` to handle the initial checks.

chrisfilo:
We are not using ModelPermissions, but that would make sense for operations such as creation of new objects. Are you basically saying that ObjectPermissions work only if ModelPermissions are used? It might be worthwhile to start using ModelPermissions - this might be a cleaner solutions. If I understand things correctly...

burnash:
Yes, what I'm saying is `DjangoObjectPermissions` from DRF works only if ModelPermissions are used. Otherwise it'll fail any update request to protected endpoint with `HTTP_403_FORBIDDEN` at [request dispatch stage](https://github.com/tomchristie/django-rest-framework/blob/4e5da16961c8f11d01a2f9dc475efacb29e778c7/rest_framework/views.py#L377). It fails because it doesn't find any permission for the current user for this model.

For the case without ModelPermissions, it looks like it's easy to fix, just inherit from `DjangoObjectPermissions` and override `has_permissions()` to handle missing ModelPermissions. (I'm going to try this now.)

> We are not using ModelPermissions, but that would make sense for operations such as creation of new objects.

Could you elaborate on this? What type of objects are you referring to?

chrisfilo:
I have nothing against starting to use ModelPermissions. What I meant is
that "ability to create a collection" is inherently a ModelPermission and
cannot be solved by ObjectPermissions. In short why don't we just start
using ModelPermissions?

burnash:
Yes, I was trying to figure out if there is any need for regulating model access for users in NeuroVault. 
If you don't have anything against turning on ModelPermissions, that'll be the easiest way to get DRF ObjectPermissions working.

chrisfilo:
Let's do that.

burnash:
Will you do this or I can turn it on in a separate PR?

chrisfilo:
If you don't mind doing this that would be faster. Thanks!

burnash:
I've rebased the PR and added special permission class for polymorphic models.

burnash:
I've rebased the branch over #419. Newly added [`ObjectOnlyPolymorphicPermissions`](https://github.com/burnash/NeuroVault/commit/e3929a38b296700bdf77b4b25d67be3fbdbcadc5#diff-e5b0bca0c4350630e8e4311ffcede57dR12) class came in handy with the migration of permissions to `BaseCollectionItem`.

chrisfilo:
Great! Is it ready for review/merge?

burnash:
Yes, it's ready for review & merge.

chrisfilo:
LGTM!

PR #401, opened 2015-12-26
urlquote in the API

### LINKED ISSUE #381: Filenames should be urlencoded.
Filenames of images must be url encoded. This filename contains `%`, and without urlencoding it, this breaks things:
http://neurovault.org/images/10483/

In Python, just call `urllib.quote`:
https://stackoverflow.com/questions/1695183/how-to-percent-encode-url-parameters-in-python

It's worth noting that these same unencoded URLs are returned by the API as well.

chrisfilo:
Closes #381. @bcipolli please review

### DIFF
--- a/neurovault/api/serializers.py
+++ b/neurovault/api/serializers.py
@@ -1,9 +1,8 @@
 import os
 
 import pandas as pd
-
 from django.forms.utils import ErrorDict, ErrorList
-
+from django.utils.http import urlquote
 from rest_framework import serializers
 from rest_framework.relations import PrimaryKeyRelatedField, StringRelatedField
 
@@ -22,7 +21,7 @@ class HyperlinkedFileField(serializers.FileField):
     def to_representation(self, value):
         if value:
             request = self.context.get('request', None)
-            return request.build_absolute_uri(value.url)
+            return request.build_absolute_uri(urlquote(value.url))
 
 
 class HyperlinkedRelatedURL(serializers.RelatedField):


bcipolli:
I was thinking of submitting a similar PR. Is there any chance that `value.url` contains protocol, path, or querystring information?

chrisfilo:
It's just a string. What protocol are you referring to? Does this PR solve
your problem?

bcipolli:
if `value.url` is `my_file.nii.gz`, it solves it. If it's ever `/path/to/my_file.nii.gz` or `http://server/path/to/my_file.nii.gz`, it may break.

Just wanted to check, as I don't have a local server to test it out with. Yes, I think this will fix the issue, if `value.url` is only ever a filename.

chrisfilo:
Have you tried our docker setup explained in the README? It should make
installing a dev server a breeze!

bcipolli:
Nope. Cool, I'll try it out in the morning! :+1:

bcipolli:
Installation requires much more bandwidth than I have at home, so this will have to wait until tomorrow.

bcipolli:
Unrelated to these tests: I'm able to add a collection, but keep getting 502 Bad Gateway when I add an image. I tried investingating using docker's `ubuntu` image, but to no avail. Any ideas?

image

chrisfilo:
You would have to look at the docker-compose logs (output) to debug this.

bcipolli:
Thanks. I solved that issue. I still get 404s when trying to GET the image file with images that need to be urlencoded. I am sure that I'm running your latest code.

I cannot, however, find the image files that I uploaded anywhere. What's the right command to view them inside the docker container? 

I have verified that images with typical filenames are downloadable, so I'm sure it's in the filesytem somewhere. But I can't get at it, even while manually editing the `media` download URL to every possibility I can think of.

chrisfilo:
Could you provide the steps to replicate the error you are getting?

If you have not changed anything uploaded images are stored inside the data container. You can browse them by loggin into a container:

	docker-compose run worker /bin/bash
	
Inside the container you can do:

	ls /var/www/image_data/images

chrisfilo:
I confirmed using pyneurovault that this in fact fixes the problem with invalid urls in API.

PR #404, opened 2015-12-27
Updating install instructions to use docker-machine.

bcipolli:
I was able to run each of the install instructions. The rest of the `docker-compose` commands should be valid; I'm running them now, but my home Internet is too slow for the size of this install to complete anytime soon.

Please take a look and try it out!

PR #405, opened 2015-12-27
Switch docs to docker-machine

### LINKED ISSUE #403: Migrate to `docker machine`
When running `boot2docker init`, this is splashed on my screen:
```
  WARNING: The 'boot2docker' command line interface is officially deprecated.

  Please switch to Docker Machine (https://docs.docker.com/machine/) ASAP.

  Docker Toolbox (https://docker.com/toolbox) is the recommended install method.
```

I will try getting this to work via `boot2docker`, then see if I can't get `Docker Machine` running too.

chrisfilo:
How about this @bcipolli instead? Closes #403. Closes #404.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,17 +5,17 @@ Easy to use web database for human brain statistical maps, atlases and parcellat
 ### Installing dependencies
 1. Fork the main repository (https://github.com/NeuroVault/NeuroVault)
 2. Clone your fork to your computer: `git clone https://github.com/<your_username>/NeuroVault`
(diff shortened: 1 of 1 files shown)


bcipolli:
If that's more correct, that works for me! I'm not sure that `docker-machine` is just for OSX; there were install instructions for all platforms: https://docs.docker.com/machine/install-machine/

chrisfilo:
docker-machine is only needed if you don't have Linux (otherwise you can use docker natively without the need of a virtual machine)

bcipolli:
Right; the other use-case is Windows, no? :package:

chrisfilo:
Yes  - windows potentially, but I never tested it.

PR #406, opened 2016-01-01
Protect maps when cognitive atlas contrasts or tasks are deleted.

chrisfilo:
@vsoch to avoid potential disaster


vsoch:
greatly appreciated :o)

vsoch:
He might have finally updated polymorphic models so it matches the docs:

https://django-polymorphic.readthedocs.org/en/latest/quickstart.html

I had posted about it here:

https://github.com/chrisglass/django_polymorphic/issues/180

PR #416, opened 2016-01-17
Enh/api my collections

chrisfilo:
@burnash have a look at this. I found it useful when interacting with the API.


burnash:
Looks good. I have some upcoming changes to `CollectionViewSet`: the view set will become editable. I'd like to see how it'll go with a user's own collections.

chrisfilo:
I hope those features will be inherited. Users will be able to use the new endpoint to modify their collections.

PR #417, opened 2016-01-17
Add Atlas permissions

burnash:
This PR adds object-level permissions for Atlas


burnash:
It turned out this is obsolete. The problem this PR tries to solve can't be solved this way.

PR #419, opened 2016-01-18
Make BaseCollectionItem not abstract

chrisfilo:
This PR removes the abstract nature of the BaseCollectionItem class. This cleans up quite a bit of code (including permissions) because both statistical maps and NIDMResults objects can be treated similarly. From a user point of view this PR will list NIDM Results objects on collection views along statistical maps.

All tests are currently passing apart from uploading NIDM Results files through the API. When I try to do this I get 400 and the following error:

	{"non_field_errors":["'collection'"]}
	
@burnash do you have any idea what could be causing this?


chrisfilo:
I have found and fixed the error with API NIDM files upload. Since this is a substantial refactor with rather complicated migration I would appreciate help in testing and review. @vsoch @rwblair

burnash:
@chrisfilo I didn't fully grasp all implications of this abstract to polymorphic move, but the error is indeed tells that collection field is missing and your fix is correct.

neurovault/api/serializers.py
```diff
@@ -229,8 +230,7 @@ class Meta:
 class CollectionSerializer(serializers.ModelSerializer):
     url = HyperlinkedImageURL(source='get_absolute_url', read_only=True)
     owner = serializers.ReadOnlyField(source='owner.id')
-    images = ImageSerializer(many=True, source='image_set')
-    nidm_results = NIDMResultsSerializer(many=True, source='nidmresults_set')
+    images = ImageSerializer(many=True, source='basecollectionitem_set')
```
burnash: Does it mean that `images` attribute will contain statisticmaps, nidmresults and atlases?

burnash: Is there any requirements for backward-compatibility of the API? If not one option is to introduce `items` attribute for all of BaseCollectionItem's instances.

vsoch:
I'm going to doctor first thing this morning, but you got it! Will try to look today.

chrisfilo:
Thanks! The migrations are something I am most worried about since they touch every single map in the database. I tested them (populating the database with simple collections, switching to this branch and applying the migrations), but a fresh look cannot hurt.

burnash:
Why not to duplicate the real db on a staging server and apply migrations to the real database?

neurovault/api/serializers.py
chrisfilo: Actually `items` field  of `CollectionSerializer` is overridden in `CollectionViewSet` so this change is not reflected anywhere. Maybe we should completely remove it in the future.

burnash: I think I'm missing something, because I can't see `items` field neither in `CollectionViewSet`, nor in `CollectionSerializer`.

chrisfilo: Apologies - I am experiencing some cognitive failures ;) I meant `images`.

burnash: I see. No problem. Congratulations on pulling this off!

PR #422, opened 2016-01-19
Redirect to "My collections" on Log in or Sign up

burnash:
This is a simple enhancement. There could be a different "most common user behaviour". Speaking for myself, I rarely need to edit my profile right after I log in:

screen shot 2016-01-19 at 22 48 56

A better destination for a "bare login" (i.e. without "next" url parameter) is probably "My collections".


chrisfilo:
Yes! Thank you! I was meant to do this for a while!

burnash:
I'm glad to hear you like it )

PR #425, opened 2016-01-27
Add warnings about missing metadata

chrisfilo:
When users upload folders or zip files the do not need to input metadata. However, often they are not even aware that there is anything missing. This PR adds warnings about missing metadata.

screenshot from 2016-01-27 11-52-04

### DIFF
--- a/neurovault/api/views.py
+++ b/neurovault/api/views.py
@@ -338,3 +338,4 @@ class TagViewSet(viewsets.ModelViewSet):
         from django.db.models import Count
         data = Tag.objects.annotate(action_count=Count('action'))
         return APIHelper.wrap_for_datatables(data)
+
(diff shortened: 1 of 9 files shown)


vsoch:
can the spacing of the red warning box be in line with something on the page, for the right and left?

chrisfilo:
I'm sure it can - just don't quite know how to do that. There are a few misalignments in our layout that need to be fixed.

vsoch:
Try setting padding-left / padding-right and margin-left margin-right in the style of the element, either via the warning box itself, or the layer it is sitting in.

vsoch:
It would be best to have a general warning or message class that can be used across pages instead of setting the local style. This class should be defined in a file after bootstrap, otherwise it will not override the default style.

chrisfilo:
I moved the style changes to the main CSS and monkey patched the margin. Other alignment issues will need to wait for another PR.

vsoch:
:monkey: :)

PR #445, opened 2016-02-16
Tweaks to Search and Cognitive Atlas Task View

vsoch:
This PR will do some minor tweaks on #432, specifically:

### Addition of sitemap
Django comes with [sitemaps built in](https://docs.djangoproject.com/en/1.9/ref/contrib/sitemaps/), so this was very easy to add. We basically [add a line](https://github.com/vsoch/NeuroVault/blob/enh/semantic-image-comparison/neurovault/urls.py#L36) to the urls.py to generate a sitemap.xml in the root, and the `sitemaps` variable is a dictionary that has keys as names (eg, Collections, Images, CognitiveAtlasTask) and values as SiteMap objects that are [defined here](https://github.com/vsoch/NeuroVault/blob/enh/semantic-image-comparison/neurovault/api/sitemap.py). Each class extends the django sitemaps object, and I made a BaseSitemap since there is overlap in a few things (and I added a get_absolute_url() function to CognitiveAtlasTask so we could use this in the Sitemap). The only required function is items, and this should return a query object for what we want indexed (please take a look at these filters). The location and lastmod are optional, and I included when we had that information in the models. The priority variable value indicates how we want the robot to select between choosing links when parsing a page, and doesn't have impact on the order of search results returned, and I chose 0.5 to give all objects equal priority. The changefreq represents how often the object is expected to change, and since these are not static posts, I chose "always."

The sitemap is just an xml file - and we have the choice to make different sitemap xml files for different objects, but it's only necessary when our total number of objects gets up to 50K (not even close right now). Here is what it looks like:

image

I'm not sure where "example.com" comes from, but hopefully this is returned by the development server, and will return neurovault.org in production. @chrisfilo let me know if you know anything about this.

### Changes to Cognitive Atlas Task View
The search page now performs a "get" instead of a "post" so we don't need csrf, and the url is something more easily shared (eg `http://127.0.0.1/images/task?cogatlas=trm_523c7a0a73cf5`)

The cognitive atlas task button has been added to each single image page. Since the large buttons were getting crowded and the default size on the collections pages is the normal, I changed the button sizes to be normal (eg class `btn` and not `btn-lg`) so it's still easily readable, but not so crowded:

image

I thought "Task View" was appropriate because it's short and suggests viewing something about the task.

cognitiveatlastask has been added as a meta data to the cognitive atlas task view:

image

And I fixed a bug that would trigger an error given that a user selects a task with no images. This now shows this page:

image


vsoch:
We probably would only want to include CognitiveAtlasTask that have images defined.

rwblair:
example.com is defined in the database as the default "site name":
https://docs.djangoproject.com/en/1.9/ref/contrib/sites/

Can be changed in the admin section of the site(don't think my account has admin privileges so I couldn't check to see if it was set.)

rwblair:
Looks like it is set to neurovault.org so the sitemap should work as expected.

neurovault/api/sitemap.py
```diff
@@ -0,0 +1,34 @@
...
+    def items(self):
+        return Collection.objects.filter(private=False)
+
+
+class CognitiveAtlasTaskSitemap(Sitemap):
+    changefreq = "always"
+    priority = 0.5
+
+    def items(self):
+        return CognitiveAtlasTask.objects.all()
```
chrisfilo: Can we limit this only to non-empty task categories?

chrisfilo:
Could you overload lasmod for Images and Collections to use modify_date (like in this example: https://docs.djangoproject.com/en/1.8/ref/contrib/sitemaps/). I would also change changefeq to "weekly" for Images and Collections and "daily" to Tasks. More info about these parameters can be found here: http://www.sitemaps.org/protocol.html.

vsoch:
Collections and Images extend Basesitemap so isn't that done?

chrisfilo:
I yes - sorry I missed that. 

What about changefreq? Isn't "always" too aggressive?

vsoch:
Oh yes I will fix that.. just afk atm!

vsoch:
ok, fixed! Collections and images are now weekly, and tasks daily. boombattaboom.

image

Issue #29, opened 2014-09-12
Tag searching should be case insensitive


GaelVaroquaux:
These 2 URLs should return the same thing:
http://neurovault.org/images/tags/Motor/
http://neurovault.org/images/tags/motor/

PR #450, opened 2016-02-22
Case sensitive tag search.

iCHAIT:


iCHAIT:
@chrisfilo  Hi, I have made an attempt to fix the issue, can you review and suggest if I am doing it correctly?

chrisfilo:
This would only display the tag in lower case - it does not address the search issue described in #29.

iCHAIT:
@chrisfilo Okay I'll look into it again, BTW can you tell me [how to add sample data](https://github.com/NeuroVault/NeuroVault/issues/452) to my local instance of the site?

neurovault/apps/statmaps/views.py
```diff
@@ -648,11 +648,11 @@ def delete_image(request, pk):
 
 def view_images_by_tag(request, tag):
     if request.user.is_authenticated():
-        images = Image.objects.filter(tags__name__in=[tag]).filter(
+        images = Image.objects.filter(tags__name__in=[tag.lower()]).filter(
                                         Q(collection__private=False) |
                                         Q(collection__owner=request.user))
     else:
-        images = Image.objects.filter(tags__name__in=[tag]).filter(
+        images = Image.objects.filter(tags__name__in=[tag.lower()]).filter(
```
chrisfilo: If an image is tagged with "Brain" or "BRAIN" it will not  be captured by this query (for tag=="brain").

erramuzpe:
Hi @iCHAIT  @chrisfilo 

I've been working as well in this and is not so easy...

this issue can be solved with the following custom lookups for model fields:

- [For exact non case sensitive matching: ](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#std:fieldlookup-iexact)
- [ For substring non case sensitive matching:](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#std:fieldlookup-icontains)

The problem is that  _tag_ has not this lookup options. This could be done with 'django-taggit'.



Other (not elegant) possible solutions could be:

1. When saving the tags, save all of them in lowercase and then when making the query pass the tag as lowercase too. 
2. Generating [all possible capitalization combinations](http://stackoverflow.com/questions/6747123/combination-of-all-possible-cases-of-a-string)
(this will generate problems for long tag searches if it is not controlled, say [if len(tag)>10;print 'tag too long']. A tag of 10 chars will generate a list of 1024 combinations, which makes no sense...). 


The following code actually works to solve the problem, but I want to hear from you.
```
import itertools
def cc(s):
	return (''.join(t) for t in itertools.product(*zip(s.lower(), s.upper())))

def view_images_by_tag(request, tag):
    if request.user.is_authenticated():
        images = Image.objects.filter(tags__name__in=list(cc(tag))).filter(
                                        Q(collection__private=False) |
                                        Q(collection__owner=request.user))
    else:
        images = Image.objects.filter(tags__name__in=list(cc(tag))).filter(
                                        collection__private=False)
    context = {'images': images, 'tag': tag}
    return render(request, 'statmaps/images_by_tag.html.haml', context)
```

 Also, it would be nice to add .distinct() at the end of the lines to avoid repeated results, if needed.

Any other ideas?

chrisfilo:
But we are using `django-taggit`...I has a config option for case insensitive searches: http://django-taggit.readthedocs.org/en/latest/getting_started.html Maybe all it takes to switch it on?

erramuzpe:
Aw.. sorry, didn't check! Switching on in my local does not solve the problem. 

It shows a strange behavior with previously created tags. So, for example, if we had previously these tags> 'Motor' and 'motor'
and we try to create a new tag 'motor' in any other image it will show the following error:

get() returned more than one KeyValueTag -- it returned 2!

Still is not solving the problem of tag searching, because it will map all new tags with case insensitiveness to the firstly created tag, and that will be the unique tag working for the search. So, say you create a tag for first time whit name 'bRaIn', and them you add another tag in another volume (or even changing 'bRaIn') whit name 'brain'. This will look for previously created tags with case insensitiveness (this is the reason of the error with previously created tags), and will return 'bRaIn', mapping the new tag to it. Then, when searching, only 'bRaIn' will return matches. 

What about lowercasing all the tags in the DB and forcing always tags to be lowercase?

chrisfilo:
It seems that this may be solved by a migration lowercasing existing tags and a changing django-taggit setting to be case insensitive. It might be worth looking into django-taggit issues to see if other users run into similar situation (switching on case insensitivity after creating bunch of tags).

erramuzpe:
An issue was opened and some of them recommend to lowercase the db:
https://github.com/alex/django-taggit/issues/308

This guy implemented management commands to do so, I've tried them and they didn't work for me (maybe doing something wrong).
http://yuguangzhang.com/blog/merge-tags-with-django-taggit/

erramuzpe:
I am still trying to solve this. With the `TAGGIT_CASE_INSENSITIVE = True` from the beginning it works: all the new tags are forced lowercase and there is no problem. With previously tagged cases, I'm having troubles when trying to lowercase the tags; 

```
value_tags = ValueTaggedItem.objects.all().distinct('object_id')
for i, value in enumerate(value_tags):
    image = Image.objects.all().filter(id=value.object_id)
    tags = image[0].tags.all()
        for i, tag in enumerate(tags):
            tag.name = tag.name.lower()
            tag.save()
```

_IntegrityError: duplicate key value violates unique constraint "statmaps_keyvaluetag_name_key"
DETAIL:  Key (name)=(motor) already exists._

https://github.com/erramuzpe/NeuroVault/blob/f14d78cff8ed64548c8b76004084c50164248dd8/neurovault/apps/statmaps/models.py#L268-L273

Any clues?

iCHAIT:
As I am not able to find time to fix this I'll close this PR for now so that it does not stop others from opening a PR.

Issue #458, opened 2016-02-25
Empty tasks should redirect or prevent indexing in other way


chrisfilo:
Right now empty tasks are listed as "Search" which is misleading.

vsoch:
It's likely they belong to private collections - we don't filter for that here:

https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/api/sitemap.py#L31

but we do for the view

https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/views.py#L414

chrisfilo:
Could be - we should exclude private collections.

avisrivastava254084:
Hi Gaffer!
Just sent you a mail. Please go through it, the subject of which is subjected as "Empty tasks should redirect or prevent indexing in other way".
Thanks.

chrisfilo:
Hi @avisrivastava254084! Let's keep the conversation here. Fixing this bug involves the following:

- [ ] Change what is displayed when a task id with no images is requested (such as http://neurovault.org/tasks/trm_55a6a8e81b7f4). It should be simple modification of the normal task page, but with text explaining that there are no maps in neurovault tagged with this task.
- [ ] The page should also have a <meta name="robots" content="noindex"> in the head so it will not show up on google search.

avisrivastava254084:
Dear Sir,
As I have seen #458 has been closed, which one would be better(better for a
Python/Django developer) to contribute to among the following:
1. #455
2. #436
Thanks.

On Wed, Mar 2, 2016 at 6:05 AM, Chris Filo Gorgolewski <
notifications@github.com> wrote:

> Closed #458 <https://github.com/NeuroVault/NeuroVault/issues/458>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/NeuroVault/NeuroVault/issues/458#event-573941826>.
>



-- 
Aviral Srivastava
+918004860407

chrisfilo:
#436 has been resolved by a recent merge. #455 is a good candidate.

Issue #430, opened 2016-02-13
Better search interface


vsoch:
We need a cohesive, single place that serves the following purposes:

- general search across fields for all statistical maps
- search based on cognitive atlas task (it would be ideal to have concept, but we don't have that level of detail represented in the database
- along with this, a view based on cognitive atlas task, and it would be nice to have a button to easily download all images (for meta analysis, etc.)

This combines the following (old) issues:

#247 
#228 
#217 

It makes sense to first work on the task specific views, and then integrate into a search framework to improve upon what we currently have.

chrisfilo:
+1

chrisfilo:
On the topic of concepts - can we get this via Cognitive Atlas API? A quick call to the API asking for all tasks/contrasts for a given concept? Then query NeuroVault for those tasks/contrasts?

vsoch:
Yes we definitely can. I think it would be more useful to have a concepts view, because people tend to want to understand images in that context. Is it worth making a task / contrast view, or should we just do concept?

chrisfilo:
I would start with Task since it does not require changes to cogatlas API
and provides some useful information. It could be easily expanded to
concepts later (the queryset would just change from asking for one task to
asking for a set of task/contrasts).

vsoch:
ok

vsoch:
Quick update: task view is traditional viewer:

![image](https://cloud.githubusercontent.com/assets/814322/13030642/92359fe6-d266-11e5-86b4-402040f923ac.png)

images are retrieved with json call as with collections, url is like:

      /images/trm_523c7a0a73cf5/task

The second view is to view the graph of the task:

![image](https://cloud.githubusercontent.com/assets/814322/13030664/0900696c-d267-11e5-8396-03a5faebd3c0.png)

This shows the task --> contrasts --> concepts, and I'm going to try and do the following:

- give meaningful colors (and legend)
- provide images as nodes at end, so user can click and go to image page

If there are no images matching the task:

![image](https://cloud.githubusercontent.com/assets/814322/13030653/cc6d116c-d266-11e5-939c-7a8660470045.png)

It might also be nice to have a button to download the image set, and some other random stuffs. Still working!

vsoch:
Colors and legend

![image](https://cloud.githubusercontent.com/assets/814322/13030777/3cf9ebf4-d26b-11e5-8782-70de3aa44f1a.png)

vsoch:
[...]

PR #461, opened 2016-03-01
Tweaks to Search and Cognitive Atlas Task View

vsoch:
This PR will do the following:

### Search Box Removed
remove explicit task search box - it's now just a row --> column of width 12:

image

### GET requests removed
remove code for handling GET requests (we are not using them anymore)

- GET is removed from [main function to view a task](https://github.com/NeuroVault/NeuroVault/compare/master...vsoch:fix/search?expand=1#diff-7129b607d02d4a1b46380abd3fd7d834R406)

This also means [changing the url](https://github.com/NeuroVault/NeuroVault/compare/master...vsoch:fix/search?expand=1#diff-011c1a7abfcd5bcca6590629636105baR114) on the statisticmaps pages (no other pages right?)

### View Task Function Changes
The new flow of the view_task function is as follows:

1) if user navigated to neurovault.org/tasks and there is no ID specified, we show them this:

image

If an ID is given, we then look for the task id in the database. If it doesn't exist, we show them this: 

image

If the ID is valid, we now have two options. We either have images tagged in the database, or not. If we do:

image

If we do not:

image

If we look at the header of this page, it will not be indexed by robots:

image

I left the other meta tags, since it doesn't hurt to still have them, but if it might be better to remove them, that's an option. I didn't see any harm in keeping them. If the task does have images, we do index:

image

### DIFF
--- a/neurovault/apps/main/templates/base.html
+++ b/neurovault/apps/main/templates/base.html
@@ -64,7 +64,7 @@
         <div class="collapse navbar-collapse" id="main-links">
             <ul class="nav navbar-nav">
                 <li class="dropdown">
-                    <a href="" class="dropdown-toggle" data-toggle="dropdown">
+                    <a href="" class="dropdown-toggle" id="collections_nav" data-toggle="dropdown">
                         Collections
                         <b class="caret"></b>
                     </a>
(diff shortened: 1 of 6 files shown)


chrisfilo:
Great job! Closes #458 #430 #247

Issue #352, opened 2015-10-18
Make lists of fields a bit longer


GaelVaroquaux:
It seems that fields are by default limited to listing 10 entries. In many situations, a longer list by default would require less clicks from the user. I have in mind in particular images, as in
http://neurovault.org/images/10480/
or the collection page
http://neurovault.org/collections/16/

For these (and maybe others), I would suggest defaulting to 25 entries in the list.

souravsingh:
I would like to work on the issue.How do I start?

chrisfilo:
The only problem I have with longer list of images on the collection view is that it will make the view asymmetrical (height of the Papaya viewer will be shorter than list of images). Any ideas how to make it look better? @erramuzpe

erramuzpe:
Yes, I realized about that (that's the reason why it was 7 by default). I don't know what to say, not really a design expert :) I would sacrifice the design in favor of usability until something is figured out.

chrisfilo:
Yes let's do that (7 with viewer, 25 without).

PR #469, opened 2016-03-13
Changed default list size for collections, images, collection index and tag listing to 25.

erramuzpe:
This solves issue #352 , but can be discussed if it's appropriate to show that fixed amount of images, tags, collections. Now it is easy to change this default setting. Also, there is the option of showing 'all' instances, in case you thing that would be interesting (it is a bit dangerous too, since it could overload the db when the project grows).

### DIFF
--- a/neurovault/api/serializers.py
+++ b/neurovault/api/serializers.py
@@ -265,7 +265,11 @@ class CollectionSerializer(serializers.ModelSerializer):
     images = ImageSerializer(many=True, source='basecollectionitem_set')
     contributors = SerializedContributors(required=False)
     owner_name = serializers.SerializerMethodField()
+    number_of_images = serializers.SerializerMethodField('num_im')
 
+    def num_im(self, obj):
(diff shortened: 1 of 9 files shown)


erramuzpe:
Seems like Docker Engine should be updated.

chrisfilo:
I was - please rebase/merge :)

erramuzpe:
This PR depends on what we agree in the discussion of issue #352

erramuzpe:
@chrisfilo I changed it according to what was agreed, but it is my first time pushing to a PR that was previuosly created and also there have been changes in Master, I'm sorry about the mess. Is it everything ok? (or at least a bit...)

chrisfilo:
Looks ok, but when in doubt rebase/merge.

erramuzpe:
I feel like: 
https://i.imgur.com/XYLwHpM.webm (PR! oh wait, rebase/merge!)

Issue #337, opened 2015-09-10
Collections API does not include number of images


vsoch:
This would be very useful, if not necessary :)

PR #470, opened 2016-03-14
Added number of images to Collections API, this solves issue #337

erramuzpe:
Solves #337. It is possible to change the position if needed. Something is wrong with the docker-compose v2 file :S (way slower than the previous one).


chrisfilo:
The slowness of the CircleCI comes from the fact that after the upgrade the cache had to be regenerated (the docker image was not reused, but it was rebuilt from scratch). Subsequent test should be as quick as before.

Issue #462, opened 2016-03-09
URL should be updated with appropriate "?q=..." after each search


chrisgorgo:
This way if someone searches for a term and presses "Back" they will be taken to the same results page.

PR #471, opened 2016-03-15
Search - Back

greyshadows:
#462 

Without changing anything in the URL. Using the history. I guess it should work.

### DIFF
--- a/neurovault/api/serializers.py
+++ b/neurovault/api/serializers.py
@@ -265,7 +265,11 @@ class CollectionSerializer(serializers.ModelSerializer):
     images = ImageSerializer(many=True, source='basecollectionitem_set')
     contributors = SerializedContributors(required=False)
     owner_name = serializers.SerializerMethodField()
+    number_of_images = serializers.SerializerMethodField('num_im')
 
+    def num_im(self, obj):
+        return obj.basecollectionitem_set.count()
+        
     def get_owner_name(self, obj):
         return obj.owner.username
 


chrisfilo:
A few things:
- Please amend the commit so it would not include the .nii.gz files and push force. This will avoid blowing up our repo size.
- This solution does not solve anything. When you run `window.history.go(-1)` you will be taken to the previous site, not the previous search. Furthermore, if you 1) search for a term 2) click a link to an image 3) press "back" in your browser you will end up on an empty search page instead of the search page for your term. Imagine how annoying it would be if Google worked this way!

The solution lies somewhere in google custom search documentation. It might require some custom javascript.

chrisfilo:
You can try using "setSearchCompleteCallback" (https://developers.google.com/custom-search/docs/js/cselement-reference#customsearchcontrol-methods) and append ?q=<search term> to the URL (see http://stackoverflow.com/questions/824349/modify-the-url-without-reloading-the-page)

vsoch:
I'm not sure if this is the right PR for this, but it might be useful long term to keep a record of exactly what people are searching for. I haven't looked into it (it may be possible in the admin tools for the search) but if not, would be worth to save some kind of log on our end.

chrisfilo:
Indeed a bit of an offtopic, but all searches are recorded together with
the rest of google analytic information.

greyshadows:
@chrisfilo  1. amend done. Pushed force. 2. Thanks for the reference. I'll look into it.

Issue #420, opened 2016-01-19
Add a close account feature to Settings


burnash:
Not the favourite feature of any web-app developer, but:

* It's polite.
* From time to time I create new NeuroVault accounts for manual testing purpose and it'd be great if I have a way to discard these accounts.

souravsingh:
I would like to help in this issue.How do I start?

onyekaa:
Saw this issue in the django group and wondered if you've resolved this...if not. What do you want to happen when an account is closed?

chrisfilo:
This has been closed by #472

PR #472, opened 2016-03-16
Delete user for issue #420

erramuzpe:
Added a new tab in the user control for account deletion, includes double check. Firstly I just did put a button and when pressed, the user and all his data was deleted. Then I realized that this is too dangerous, so I added a GitHub-like checking; the user should write his own username in order to delete the account. 
This approach is just my solution, this "deletion" stuff must be taken with care, so maybe you would like to do it in a different way.


erramuzpe:
Ah! While testing the behaviour of this PR, I realized that 2 different users can have the same email address, is this desirable?

chrisfilo:
This is true, but it has not been a problem so far.

Issue #351, opened 2015-10-15
add a share by email button


chrisfilo:

souravsingh:
Do we need to create wrapper view around django.core.email or we need to make an app from scratch?

souravsingh:
Since I believe we can create a a view around the email module to enable sharing

greyshadows:
@chrisfilo Is not present at the moment ?
http://neurovault.org/ I can see share by email below.

chrisfilo:
The idea is to extend this to collection and image views.

PR #477, opened 2016-03-23
Few Fixes

Zzoey:
Fixes for  #462 #351 #482 #479  ! :)

### DIFF
--- a/neurovault/apps/statmaps/templates/statmaps/my_collections.html.haml
+++ b/neurovault/apps/statmaps/templates/statmaps/my_collections.html.haml
@@ -40,4 +40,16 @@
 
 .row
 	%a.btn.btn-primary{href: "{% url 'new_collection' %}" } Add new collection
+
+.row
+    .span11.text-center
+    	.massive
+			<div class="addthis_toolbox addthis_default_style" style="width: 400px; margin: 0 auto;">
+			<a class="addthis_button_facebook_like" fb:like:layout="button_count"></a>
+			<a class="addthis_button_tweet"></a>
+			<a class="addthis_button_google_plusone" g:plusone:size="medium"></a>
+			<a class="addthis_counter addthis_pill_style"></a>
+			</div>
+			<script type="text/javascript" src="//s7.addthis.com/js/300/addthis_widget.js#pubid=xa-51b9d30552a8a690"></script>
+
 {% endblock %}
--- a/neurovault/apps/statmaps/templates/statmaps/search.html
+++ b/neurovault/apps/statmaps/templates/statmaps/search.html
@@ -3,7 +3,7 @@
 {% block head %}
 <script src="{% static "scripts/jquery-1.10.2.min.js"%}" type="text/javascript"></script>
 <script src="{% static "scripts/select2.min.js"%}" type="text/javascript"></script>
-<link href="{% static "css/select2.min.css"%}" rel="stylesheet"></script>
+<link href="{% static "css/select2.min.css"%}" rel="stylesheet">
 <script src="{% static "scripts/bootstrap.min.js"%}" type="text/javascript"></script>
 <link href="http://netdna.bootstrapcdn.com/font-awesome/3.1.1/css/font-awesome.css" rel="stylesheet">
 <title>{% block title %}Search{% endblock %}</title>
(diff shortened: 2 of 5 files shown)


neurovault/apps/statmaps/templates/statmaps/my_collections.html.haml
```diff
@@ -40,4 +40,16 @@
...
+    .span11.text-center
+    	.massive
+			<div class="addthis_toolbox addthis_default_style" style="width: 400px; margin: 0 auto;">
+			<a class="addthis_button_facebook_like" fb:like:layout="button_count"></a>
+			<a class="addthis_button_tweet"></a>
+			<a class="addthis_button_google_plusone" g:plusone:size="medium"></a>
+			<a class="addthis_counter addthis_pill_style"></a>
+			</div>
+			<script type="text/javascript" src="//s7.addthis.com/js/300/addthis_widget.js#pubid=xa-51b9d30552a8a690"></script>
+
```
chrisfilo: Can we use Django templates to avoid repeating this code?

neurovault/apps/statmaps/templates/statmaps/search.html
```diff
@@ -67,11 +67,15 @@
 		    var s = document.getElementsByTagName('script')[0];
 		    s.parentNode.insertBefore(gcse, s);
 		  })();
+
 		</script>
-		<gcse:search linkTarget="_self"></gcse:search>
+		<!-- <gcse:search linkTarget="_self"></gcse:search> -->
+    <gcse:searchbox-only resultsUrl="http://192.168.99.103/search" queryParameterName="as_q"></gcse:searchbox-only>
```
chrisfilo: use relative path

chrisfilo: is defining query parameter name necessary? what's wrong with the default "q"?

chrisfilo:
Please avoid sending a single PR for multiple issues in the future. This causes unnecessary dependencies - for example if the fix for the typos (missing " for example) was a separate PR it would be a no brainer to merge straight away.

neurovault/apps/statmaps/templates/statmaps/search.html
greyshadows: Yeah. It's not required. I used for reference. Changes done. (URL should be neurovault.org/search).

greyshadows:
Yeah. Sure. 

Thanks. :)

neurovault/apps/statmaps/templates/statmaps/search.html
```diff
@@ -70,8 +70,8 @@
 
 		</script>
 		<!-- <gcse:search linkTarget="_self"></gcse:search> -->
-    <gcse:searchbox-only resultsUrl="http://192.168.99.103/search" queryParameterName="as_q"></gcse:searchbox-only>
-    <gcse:searchresults-only queryParameterName="as_q" linkTarget="_self"></gcse:searchresults-only>
+    <gcse:searchbox-only resultsUrl="http://neurovault.org/search"></gcse:searchbox-only>
```
chrisfilo: Is there a way to use a **relative** URL so it would work on both production and development?

neurovault/apps/statmaps/templates/statmaps/my_collections.html.haml
greyshadows: Yes. I can do that.

neurovault/apps/statmaps/templates/statmaps/search.html
greyshadows: I don't think so. There is an other way to do this but it is indirect. This is the best way possible from the documentation of CSE. :) I will try looking for it.

neurovault/apps/statmaps/templates/statmaps/search.html
```diff
@@ -67,11 +67,15 @@
 		    var s = document.getElementsByTagName('script')[0];
 		    s.parentNode.insertBefore(gcse, s);
 		  })();
+
 		</script>
-		<gcse:search linkTarget="_self"></gcse:search>
+		<!-- <gcse:search linkTarget="_self"></gcse:search> -->
```
chrisfilo: remove the comment

neurovault/apps/statmaps/templates/statmaps/search.html
```diff
@@ -67,11 +67,15 @@
...
+    <gcse:searchbox-only resultsUrl="http://neurovault.org/search"></gcse:searchbox-only>
```
chrisfilo: if you use:

    <gcse:searchbox-only resultsUrl="search"></gcse:searchbox-only>

it will work in both production and development (this is what I meant by a relative URL).

PR #483, opened 2016-03-29
Deleted lines already satisfied

erramuzpe:
Irrelevant PR  ¯\_(ツ)_/¯
sorry (¬_¬)


erramuzpe:
IDK: What is the point of installing all pip dependencies in the Dockerfile and then from this same file installing the requirements in requirements.txt?   

EG: in the Dockerfile -> pybraincompare==0.1.18
in requirements.txt -> pybraincompare==0.1.15

As requirements.txt comes after the "pip install" command, this will make the version 0.1.18 to be uninstalled in benefit of 0.1.15. Right? Probably I'm missing something here...

chrisfilo:
Yeah so this setup is a bit complicated. We used to rely mostly on requirements.txt, but this made rebuilding the docker image long since Docker is caching each line of Dockerfile separately. So if you install everything in one line and you modify say version of pybraincompare all python dependencies will be rebuilt (including the very time consuming compilation of numpy and scipy). The more I think about the more I lean towards dropping requirements.txt and just relying on Dockerfile. This would avoid the bug with pybraincompare version update you spotted.

erramuzpe:
So, before deleting requirements.txt I checked 1 by 1 all the dependencies.

- Updated in Dockerfile _django-oauth-toolkit==0.10.0_ (it was 0.10.0 in requirements and 0.9.0 in Dockerfile).
- Left in Dockerfile _django-file-resubmit==0.4.3_ instead of git+https://github.com/sinnwerkstatt/django-file-resubmit.git#egg=file_resubmit in requirements.txt (I saw a previous modification of you, so I guessed...).
- Left in Dockerfile the pycortex line (it was slightly different in requirements.txt).
- Deleted all quotation marks and the -v of the scipy.
- Ordered a-z
- deleted 
_ADD requirements.txt /code/ &&
RUN pip install -r requirements.txt_
but not
_RUN mkdir /code &&
WORKDIR /code_

Hope everything is fine and works for you :key:

Dockerfile
```diff
@@ -6,62 +6,57 @@ RUN apt-get update && apt-get install -y \
...
+RUN pip install Django==1.8.8
 RUN pip install djangorestframework
-RUN pip install markdown
+RUN pip install django-celery
+RUN pip install django-chosen
+RUN pip install django-cleanup==0.4.2
+RUN pip install django-coffeescript
+RUN pip install django-crispy-forms
+RUN pip install django-datatables-view
+RUN pip install django-dbbackup<2.3
```
chrisfilo: you have to put it in quotes

erramuzpe: Aw, so sorry for that... I tried a pip install without quotations and it worked, so I though it would work for all of them.

Issue #255, opened 2015-05-11
show disabled "find similar" button for private images


chrisfilo:
With a tooltip explaining that this awesome feature is available only for public collections. We have to fix #131 first

PR #496, opened 2016-05-02
[WIP] Populating db issue85

erramuzpe:
**This PR must be coordinated with the one in Neurovault/neurovault_data** https://github.com/NeuroVault/neurovault_data/pull/2

Instructions from @jbwexler were:

Instructions:

1. ./manage.py download_fixtures
2. ./manage.py loaddata users1 statmaps1  (just "dumpdata" in this case)
3. ./manage.py collectmedia --noinput

arguments also be added to step 1 to download specific fixtures:
./manage.py download_fixtures users1 statmaps1
I've been using just a whole fixture (not users and statmaps separated).

So, there are some things to solve:

- 1 fixture or many of them (users, statmaps... separated)? Capability of downloading just some of them or everything together?
- This sample data (1 study, 1 MRI set and 1 NIDM)? We should add more for the default?
- Command based (optional for the user) or populate dumpdata by default when "docker-compose up"?


chrisfilo:
Please rebase/merge master to make this error go away.

neurovault/apps/statmaps/management/commands/download_fixtures.py
```diff
@@ -0,0 +1,59 @@
...
+                    # extract everything except for _MACOSX folder, which contains redundant data
+                    z.extractall(temp1, [x for x in z.namelist() if not x.startswith('__MACOSX')])
+            for folder in os.listdir(temp1):
+                copyTree(os.path.join(temp1, folder), fixturesDir)
+            shutil.rmtree(temp1)
+
+        else:
+            temp1 = tempfile.mkdtemp()
+            temp2 = tempfile.mkdtemp()
+            command = "svn export https://github.com/erramuzpe/neurovault_data/trunk/fixtures %s --force" % temp1
```
chrisfilo: can we avoid the SVN requirement?

chrisfilo:
To answer your questions:

1. one fixture seems easier for most usecases
2. it would be good to have at least two collections with multiple maps each, at least one of them should have a DOI attached (so it would be displayed on the main page, and similarity would be calculated for the current implementation of image search)
3. maybe we could have a command, but in the startup script check if the database is populated - if not load the fixture

neurovault/apps/statmaps/management/commands/download_fixtures.py
chrisfilo: BTW - did you know about this? http://stackoverflow.com/questions/2751227/how-to-download-source-in-zip-format-from-github/34174216#34174216

erramuzpe:
Ok! I'm into it

chrisfilo:
The good news is that it works! Some things are missing though:

- "Find similar" functionality does not work on the images added through the fixture. "Reduced representation" files .npy are missing.

- Seems like the thubnails (glass brain jpg) are included in the fixture. Could we generate them during import? This would be more robust (imaging we change the thubnail generation methods in the future - we will have to recreate the fixture).

- I'm gettin a lot of  `uwsgi_1   | File path (/code/neurovault/apps/statmaps/tasks.py\) found in fixture but not on disk in (/code/neurovault/apps/statmaps/tasks.py\)` errors - any thought where they might be coming from?

- Could you add to README.md info about the initial username and password and add another .md file explainign how to add new data to fixtures?

- Data from fixtures seems to be missing matadata. Could we include those in the import?

- Maps in NIDM collection do not seem to be in MNI - could we use some other data?

I hope this helps!

erramuzpe:
Nice!!

_"Find similar" functionality does not work..._

I will work on that, no idea this was happening

_Seems like the thubnails (glass brain jpg) are included in the fixture..._

Yes, I know about this, and you are right. They were generated in some of my tests and I copied them, then I was a bit lazy about not to deleting them... 

_I'm gettin a lot of uwsgi_1 ... errors - any thought where they might be coming from?_

I am aware of this, I will take a look 'cause it is a thing related to media-fixtures loading. 

_Could you add to README.md info ... how to add new data to fixtures?_

Sure thing, I will do it when we decide that everything is OK to merge. ok?

_Data from fixtures seems to be missing matadata. Could we include those in the import?_

Sure thing!

_Maps in NIDM collection do not seem to be in MNI - could we use some other data?_

Yes, I will download some from "real" Neurovault. I was just using some data I found from a previous PR. **This is related to the previous question**: Do you prefer some specific dataset to be the ones loaded? 

_I hope this helps!_

Nice, sure it does :+1:

chrisfilo:
I don't have any preference towards a particular collection. Maybe grab something published that looks reasonable - like this one: http://neurovault.org/collections/866/

erramuzpe:
I've solved everything except for the find_similar feature. I've been trying long to understand why it was not working... no idea. Finally I took another collection, the one proposed was too big.

erramuzpe:
Maybe related to #255 ? The only thing I can think of is that I deleted some lines in the fixture related to permissions that I deleted due to problems importing the fixture and this could cause "find similar" to be disabled.

erramuzpe:
Indeed, I changed the permissions of collections from Public to Private and Public again and now "find similar" is enabled. I'm trying to solve this, but maybe it won't be straight forward

chrisfilo:
Changing from Private to Public triggers regeneration of thumbnails and comparisons here: https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L165.

However, this should be triggered when the images are saved initially: https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L511. Maybe loading fixtures does not trigger the save() method?

erramuzpe:
Oh, this is what is happening! There is no call to save(), so this is not
being generated
El 19/05/2016 17:33, "Chris Filo Gorgolewski" <notifications@github.com>
escribió:

Changing from Private to Public triggers regeneration of thumbnails and
comparisons here:
https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L165
.

However, this should be triggered when the images are saved initially:
https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L511.
Maybe loading fixtures does not trigger the save() method?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub
<https://github.com/NeuroVault/NeuroVault/pull/496#issuecomment-220362527>

erramuzpe:
In principle, all the issues solved. I will need to change the "download_fixtures command" after merging the PR in neurovault_data

chrisfilo:
Some things I found:

1. The "atlas" collection has only parcellations, but no real atlases (atlas is a parcellation with label file in .xml format see https://github.com/NeuroVault/NeuroVault/tree/master/neurovault/apps/statmaps/tests/test_data/api
2. "NIDM test set 2" link to paper does not work and is missing authors.
3. "NIDM test set 2" does not include any NIDM files - maybe you could rename it to something else?

Otherwise it seems good!

erramuzpe:
1.- I'll upload a real atlas to that collection right now
2.- Oh, yes, I put an invented DOI, as it was a test-set. I'll change it to a real one with real authors, but it won't match with the real dataset. Is it ok?
3.- Yes, I will rename it, no probs.

erramuzpe:
I will change the NIDM 2 dataset entirely to match a real one.

erramuzpe:
Solved all these minors. All this changes were applied in neurovault_data. When that PR is merged, there are minor changes to be applied to download_fixtures.py.

PR #500, opened 2016-05-05
tag sensitivity issue #29

erramuzpe:
This command lowercases all tags in the db. Then, there is a choice to make. 

-Force lowercase for all new tags manually (easy way). 
-`TAGGIT_CASE_INSENSITIVE=True` in settings.py (problematic way). Why? 'Cause all the Tags that were created with case differences (say 'motor' and 'Motor') have different tag-PK (KeyValueTag). This means that there is a conflict when creating a new 'motor' tag, since the db will find 2 keys for that tag and an error arises. This can be solved with a mergetags() function iterating over all conflicting Tags and deleting their PK and pointing to the lowercased one.


erramuzpe:
I resolved in the "problematic way" (it was not so problematic at the end of the day).
Way of using: 

docker exec -ti neurovault_uwsgi_1 /bin/bash
./manage.py lowercasetags
./manage.py dlduptags

and activate `TAGGIT_CASE_INSENSITIVE=True` in case it is not yet in settings.py, then, everything should work fine. 

The checking says that the lowercasing is too complex; but I had to check all the tagged images, then iterate over them, then iterate over the tags to lowercase them... Any ideas to improve it?

chrisfilo:
Can this be done with a data migration instead of a Command? It would be cleaner.

https://docs.djangoproject.com/en/1.9/topics/migrations/#data-migrations

erramuzpe:
When running the code in the Django shell, it works; when running `./manage.py migrate` from the naurovault image bash, it does not. The problem is that when querying for objects, say...

```
Image = apps.get_model("statmaps", "Image")
image = Image.objects.all()
```

In the Django shell I get "[<StatisticMap: zstat old young>, <StatisticMap: zstat old>, <StatisticMap: zstat young old>, <StatisticMap: zstat young>, <StatisticMap: MNI maxprob thr50 2mm>, <StatisticMap: MNI maxprob thr0 1mm> ... "

But in the migration script seems not to be working in the same way, I get: 
[<Image: Image object>, <Image: Image object>, <Image: Image object>, <Image: Image object>, <Image: Image object>, <Image: Image object>, <Image: Image object>, <Image: Image object> ... ]

and then, of course, the attribute managing is different. Why is this happening?

chrisfilo:
I guess this has something to do with Django Polymorphic. Could you send a PR a separate PR with Django Polymorphic version bump. I'll merge it straight away.

erramuzpe:
I did rebuild the docker image and still having the same issue, will take a look on it tomorrow and try to solve in the migration script :+1:

chrisfilo:
Everything works as advertised! Is it still work in progress or can I merge it?

erramuzpe:
No more WIP, if it works for you, then is time to merge :dancer:

PR #504, opened 2016-05-12
Neurovault - Neuropower integration

jokedurnez:
There are two main parts of this pull request:
1. A button is added to the image pages on [neurovault](http://www.neurovault.org).  This button redirects to [neuropower](http://www.neuropowertools.org) and fills out the form with data from the neurovault API, similar to the [neurosynth](http://www.neurosynth.org) button.
2. A specific data upload page has been created on [neurovault](http://www.neurovault.org) for power analysis.  This page allows with one page the upload of a new image, also similar to what is in [neurosynth](http://www.neurosynth.org).  The form is specific to the necessary input for neuropower.  Images are by default hidden, as these data are supposed to be from pilot or incomplete studies.

### DIFF
--- a/neurovault/apps/statmaps/views.py
+++ b/neurovault/apps/statmaps/views.py
@@ -37,7 +37,7 @@ from xml.dom import minidom
 
 import neurovault
 from neurovault import settings
-from neurovault.apps.statmaps.forms import CollectionForm, UploadFileForm, SimplifiedStatisticMapForm,\
+from neurovault.apps.statmaps.forms import CollectionForm, UploadFileForm, SimplifiedStatisticMapForm,NeuropowerStatisticMapForm,\
     StatisticMapForm, EditStatisticMapForm, OwnerCollectionForm, EditAtlasForm, AtlasForm, \
     EditNIDMResultStatisticMapForm, NIDMResultsForm, NIDMViewForm, AddStatisticMapForm
 from neurovault.apps.statmaps.models import Collection, Image, Atlas, StatisticMap, NIDMResults, NIDMResultStatisticMap, \
@@ -272,7 +272,7 @@ def view_image(request, pk, collection_cid=None):
 
 
 def view_collection(request, cid):
-    '''view_collection returns main view to see an entire collection of images, meaning a viewer and list of images to load into it. 
+    '''view_collection returns main view to see an entire collection of images, meaning a viewer and list of images to load into it.
     :param cid: statmaps.models.Collection.pk the primary key of the collection
     '''
     collection = get_collection(cid,request)
@@ -413,7 +413,7 @@ def view_task(request, cog_atlas_id=None):
     if task:
         images = StatisticMap.objects.filter(cognitive_paradigm_cogatlas=cog_atlas_id,
                                              collection__private=False).order_by("pk")
-        
+
         if len(images) > 0:
             first_image = images[0]
             graph = get_task_graph(cog_atlas_id, images=images)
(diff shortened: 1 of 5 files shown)


chrisfilo:
Looks good - couple of things:

- It might be worth making "number_of_subjects" required in your version of the form. See here how to do this: http://stackoverflow.com/questions/21343219/django-form-fields-change-to-mandatory

- Codacy picked on the fact that there is a lot of repeated code in `add_image_for_neuropower` and `add_image_for_neurosynth`. Could you extrat it into a single function with following parameters: template path, redirect URL template, and is_private? This  new function could be called in both  `add_image_for_neuropower` and in `add_image_for_neurosynth`. This way we will reduce code redundancy.

jokedurnez:
Hmm...  The tests are better, I'm not sure if this is what you mean.  Imo it's a bit messier...

neurovault/apps/statmaps/views.py
```diff
@@ -447,23 +444,40 @@ def add_image_for_neurosynth(request):
...
-        form = SimplifiedStatisticMapForm(request.POST, request.FILES, instance=image, user=request.user)
+        if app == "neurosynth":
+            form = SimplifiedStatisticMapForm(request.POST, request.FILES, instance=image, user=request.user)
+        elif app == "neuropower":
+            form = NeuropowerStatisticMapForm(request.POST, request.FILES, instance=image, user=request.user)
         if form.is_valid():
             image = form.save()
-            return HttpResponseRedirect("http://neurosynth.org/decode/?neurovault=%s-%s" % (
-                temp_collection.private_token,image.id))
+            return HttpResponseRedirect(redirect)
```
chrisfilo: don't you need to pass private_token and image id to the redirect URL?

neurovault/apps/statmaps/views.py
```diff
@@ -447,51 +444,40 @@ def add_image_for_neurosynth(request):
         priv_token = generate_url_token()
         temp_collection = Collection(name=temp_collection_name,
                                      owner=request.user,
-                                     private=False,
+                                     private=is_private,
                                      private_token=priv_token)
         temp_collection.save()
     image = StatisticMap(collection=temp_collection)
+    redirect = redirect_url+"/?neurovault=%s-%s" % (
+        temp_collection.private_token,image.id)
```
jokedurnez: private_token and image.id are defined here.

chrisfilo: Ah I see! Could you take it outside of the function? So the string passed on would be "http://neurosynth.org/decode/?neurovault=%s-%s" instead of "http://neurosynth.org/decode/". It will give more flexibility for future apps using this functionality.

neurovault/apps/statmaps/views.py
```diff
@@ -447,23 +444,40 @@ def add_image_for_neurosynth(request):
...
     if request.method == "POST":
-        form = SimplifiedStatisticMapForm(request.POST, request.FILES, instance=image, user=request.user)
+        if app == "neurosynth":
+            form = SimplifiedStatisticMapForm(request.POST, request.FILES, instance=image, user=request.user)
```
chrisfilo: Instead of passing "app" argument and doing this conditional could you just pass the form class (for example `SimplifiedStatisticMapForm`)?

jokedurnez:
That would mean splitting up the functions, because the temp-collection should be made before defining the url, and then you can pass the url to the final function.  The structure would be as follows:

def add_temp_collection(request, is_private):
    *** function to make temporary collection ***

def redirect(request, temp, template_path, redirect_url):
    *** function that redirects to external page ***

def add_image_for_***(request):
   temp,image = add_temp_collection(**)
   redirect_url = #define url based on temp
   return redirect(**)

I tried this at first but there are quite a few repetitions there as well: every time calling image,...  You agree with this structure?

neurovault/apps/statmaps/views.py
jokedurnez: ha, didn't know that was possible.  That'll make it definitely better.

jokedurnez:
Ha ok I see what you mean. Didn't know that trick, cool!  I tested it and it works.

chrisfilo:
LGTM - is neuropower ready to accept those requests? If so I'll merge it and deploy it.

jokedurnez:
Neuropower is ready !

PR #506, opened 2016-05-16
[WIP] Cleans tags (lowercases them)

erramuzpe:
This PR cleans the tags from a form, lowercasing them. I am not sure if this only applies for StatisticMapForms, I've been testing with several image types.

### DIFF
--- a/neurovault/apps/statmaps/forms.py
+++ b/neurovault/apps/statmaps/forms.py
@@ -493,6 +493,7 @@ class StatisticMapForm(ImageForm):
         django_file = cleaned_data.get("file")
 
         cleaned_data["is_valid"] = True #This will be only saved if the form will validate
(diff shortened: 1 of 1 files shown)


chrisfilo:
Is this necessary? I thought this is what the taggit setting was essentially doing.

erramuzpe:
I think it is, I've just arised the error again. The taggit setting is not distinguishing between "motor" and "Motor", but still an user can create a tag called "BLABLABLA" if "blablabla" was not already created. We must ensure that all created tags are lowercased, this also will avoid redundancy.

erramuzpe:
My question is if this form-level lowercasing must be applied for every form (NIDM, Atlas...), or just for StatisticMap. I've tried uploading NIDM and Atlases and it works for me, but I'm not completely sure, since they are being saved as StatisticMap objects ?¿ is this reasonable or am I losing something here?

PR #507, opened 2016-05-17
Changed tagcleaning to ImageForm and added in NIDMform

erramuzpe:
Changed clean_tag() to ImageForm and added into NIDMResultsForm. With this PR, we can avoid errors; probably the last migration has to be executed again, since there will be some duplicated tags due to the pair of errors generated...

While doing this, I've realized that  NIDM tags don't appear in the NIDM view and neither are being indexed by the tag_search (NIDM tags do exist, but they are not being used at all). Maybe this is due to a bad use from my side: I am uploading a " * .nidm.zip" file. This file generates a statistical map and a NIDM zip file (just this last one can be edited). Neurovault viewer complains about the "*.nidm.zip" of being a not recognized format, but if I try to open it, the NIDM viewer appears and I can edit the info (tags...) if I am the owner.  

:bulb: ?

### DIFF
--- a/neurovault/apps/statmaps/forms.py
+++ b/neurovault/apps/statmaps/forms.py
@@ -476,6 +476,7 @@ class ImageForm(ModelForm, ImageValidationMixin):
 
     def clean(self, **kwargs):
         cleaned_data = super(ImageForm, self).clean()
+        cleaned_data["tags"] = clean_tags(cleaned_data)
         return self.clean_and_validate(cleaned_data)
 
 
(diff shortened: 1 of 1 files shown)


chrisfilo:
NIDMResults are still missing some features, but uplodas should work
without errors. Could you open a separate issue wiht the NIDM related error
describing in details steps to replicate and the file you used?

erramuzpe:
OK!  There is no error when uploading the NIDM file, it uploads correctly. Then the small viewer refuses to show the nidm.zip (which probably is a normal behavoiur). The other issue it is just a matter of how the tags are managed. I didn't know that it is still developing, so we can let this for another moment if it is OK for you
nidm

PR #517, opened 2016-05-31
Benchmark commands

erramuzpe:
Some benchmark commands. I will merge all them together and allow different functionality using arguments. I will start doing this tomorrow.


neurovault/apps/statmaps/management/commands/benchmark3.py
```diff
@@ -0,0 +1,122 @@
...
+                    with t:
+                        run_voxelwise_pearson_similarity(
+                            image.pk)  # TODO: change this depending on the indexing function
+                    #print "Time taken to index", i, " images: ", t.interval
+                    index_table[i] = t.interval
+                    np.save(os.path.join(app_path, 'bench/results_index_busy'), index_table)
+
+                    t = Timer()
+                    with t:
+                        get_existing_comparisons(image.pk).extra(select={"abs_score": "abs(similarity_score)"}).order_by("-abs_score")[0:100]  # "-" indicates descending # TODO: change this depending on the indexing function
```
chrisfilo: Codacy Issue found: [Expression "get_existing_comparisons(image.pk).extra(select={'abs_score': 'abs(similarity_score)'}).order_by('-abs_score')[0:100]" is assigned to nothing](https://www.codacy.com/app/krzysztof-gorgolewski/NeuroVault/file/2662767550/issues/source?bid=2767031&fileBranchId=3330867#l116)

neurovault/apps/statmaps/management/commands/benchmark2.py
```diff
@@ -0,0 +1,122 @@
...
+                with t:
+                    run_voxelwise_pearson_similarity(
+                        image.pk)  # TODO: change this depending on the indexing function
+                #print "Time taken to index", i, " images: ", t.interval
+                index_table[i] = t.interval
+                np.save(os.path.join(app_path, 'bench/results_index_not_busy'), index_table)
+
+                t = Timer()
+                with t:
+                    get_existing_comparisons(image.pk).extra(select={"abs_score": "abs(similarity_score)"}).order_by("-abs_score")[0:100]  # "-" indicates descending # TODO: change this depending on the indexing function
```
chrisfilo: Codacy Issue found: [Expression "get_existing_comparisons(image.pk).extra(select={'abs_score': 'abs(similarity_score)'}).order_by('-abs_score')[0:100]" is assigned to nothing](https://www.codacy.com/app/krzysztof-gorgolewski/NeuroVault/file/2662767549/issues/source?bid=2767031&fileBranchId=3330867#l116)

neurovault/apps/statmaps/management/commands/benchmark.py
```diff
@@ -0,0 +1,157 @@
...
+                with t:
+                    run_voxelwise_pearson_similarity(
+                        image.pk)  # TODO: change this depending on the indexing function
+                #print "Time taken to index", i, " images: ", t.interval
+                index_table[i] = t.interval
+                np.save(os.path.join(app_path, 'bench/results_index_not_busy'), index_table)
+
+                t = Timer()
+                with t:
+                    get_existing_comparisons(image.pk).extra(select={"abs_score": "abs(similarity_score)"}).order_by("-abs_score")[0:100]  # "-" indicates descending # TODO: change this depending on the indexing function
```
chrisfilo: Codacy Issue found: [Expression "get_existing_comparisons(image.pk).extra(select={'abs_score': 'abs(similarity_score)'}).order_by('-abs_score')[0:100]" is assigned to nothing](https://www.codacy.com/app/krzysztof-gorgolewski/NeuroVault/file/2670237103/issues/source?bid=2767031&fileBranchId=3330867#l153)

erramuzpe:
The bench for the queries. I made part of it (the busy server part) yesterday in my labs machine, until Saturday I'm in a congress, so I'm doing the benchs in my laptop (queries may be a bit faster, since I've an SSD). Anyway, queries are superfast.

benchqueries

erramuzpe:
@chrisfilo So, making some test I've just realized that not just `run_voxelwise_pearson_similarity()` is triggering the comparison calculation, but `save_statmap_form()` too (this is what I was using to save the statmaps in the db). So, I need to figure out how to proceed, since I can not find where it is saved (Image.save() models.py?). WIP

chrisfilo:
As discussed offline - don't call `run_voxelwise_pearson_similarity` explicitly, just rely on [.save() ](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L466) triggering all of the indexing operations. If you implement the benchmark this way you will not have to change anything to rerun it when you introduce new indexing implementation.

erramuzpe:
Ok, thanks for reminding (and sorry for forgetting), that makes total sense.

.gitignore
```diff
@@ -11,6 +11,9 @@ neurovault/static/
 
 # test data
 neurovault/apps/statmaps/tests/test_data/feat/*
+neurovault/apps/statmaps/tests/bench/*
+neurovault/apps/statmaps/templates/statmaps/statisticmap_details.html.html
```
chrisfilo: ???

neurovault/apps/statmaps/management/commands/benchmark.py
```diff
@@ -0,0 +1,104 @@
...
+        self.interval = self.end - self.start
+        if self.verbose:
+            print('time taken: %f seconds' % self.interval)
+
+class Command(BaseCommand):
+    args = '<times_to_run>'
+    help = 'bench'
+
+    def handle(self, *args, **options):
+        # TODO: check if /bench folder exists. If not, donwload the full dataset.
```
chrisfilo: Could you add the downloader?

.gitignore
erramuzpe: Yes, sorry for that... in bench/  I'm saving the test data. And the statisticmap_details file is a bug that I reported as Issue. Pycharm keeps generating that file, so I encountered no other way of git telling me about that new file. I will delete both lines.

neurovault/apps/statmaps/management/commands/benchmark.py
```diff
@@ -0,0 +1,104 @@
...
+                        image_name=file,
+                        ignore_file_warning=True)
+                index_table[i] = t.interval
+                np.save(os.path.join(app_path, 'bench/results_index_busy'), index_table)
+
+                t = Timer()
+                with t:
+                    _dummy = get_existing_comparisons(image.pk).extra(
+                        select={"abs_score": "abs(similarity_score)"}).order_by("-abs_score")[
+                             0:100]  # "-" indicates descending # TODO: change this depending on the query function
```
chrisfilo: The way you benchmark querying right now will require you to change this line for the new implemetation. It would be more robust if you would just time how long does it take to generate `images/{pk}/find_similar` page instead.

neurovault/apps/statmaps/management/commands/benchmark.py
```diff
@@ -0,0 +1,104 @@
...
+# import matplotlib.pyplot as plt
+# import numpy as np
+# a = np.load('results_query_busy.npy')
+# b = np.load('results_query_not_busy.npy')
+# plt.plot(a,"ro",b,"bs")
+# plt.xlabel('Number of Images')
+# plt.ylabel('Seconds')
+# plt.title('Query benchmark for actual implementation')
+# plt.legend(['Busy server', 'Not busy'],loc=2)
+# plt.show()
```
chrisfilo: Either remove or uncomment.

chrisfilo: It would be nice to add linear and exponential fits to the plots.

neurovault/apps/statmaps/management/commands/benchmark.py
erramuzpe: No problem with this. The candidate images are: MNI templated, unthresholded and modality=fMRI-BOLD. Something else?


I've a non-benchmark-related "problem" with this;
Candidate images with previous constraints are like 5000. This is not a problem, but 1 collection of the candidates has like 4000 candidate images itself:

In one hand, http://neurovault.org/collections/503/ has 4,792 images, from a total of 11600 images in Neurovault, that makes almost half of the whole db. Downloading for testing purposes it's ok, but shall be this Collection taken into account when testing ANN queries? is biasing all the results. 

thanks chris

neurovault/apps/statmaps/management/commands/benchmark.py
erramuzpe: Yes, I was worried about this. I will do it in that way. thanks!

neurovault/apps/statmaps/management/commands/benchmark.py
erramuzpe: OK!

neurovault/apps/statmaps/management/commands/benchmark.py
chrisfilo: Indeed this is not perfect, but I think this is fine for our purposes.

erramuzpe: Can I use "pyneurovault" for this?

chrisfilo: You could, but downloading each map individually can be slow. What about using the data deposited in Stanford Digital Library Instead? https://purl.stanford.edu/qd938sk1220

erramuzpe: Ok, I will take a look into it and see how to manage.

chrisfilo: You can also consider uploading a tarball with benchmark maps on figshare

erramuzpe:
Issues completed. I will upload the plots (with fits) as soon as the benchmark finishes with ~200 images.

erramuzpe:
Plot for indexing:
index

I had problems with the exponential fitting, (this)[http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.curve_fit.html] was not working for me... ?¿ anyway, I don't think exponential fitting makes much sense with this kind of growing shape, don't you think? :thought_balloon: 


Plot for querying:
query

Since we are now querying how much find_similar takes to return the page, and the page is limited to 100 similar maps, the benchmark seems to stop growing at 100. I fitted the line to the 1st 100 results. Actually, what can be happening is that the query itself is not significant wrt the time django needs to create the html, so it could be that the real fitting should have been done from the 101 value on:
 
querying2

chrisfilo:
Aha! It's probably because find_similar limits results to the first 100! https://github.com/NeuroVault/NeuroVault/blob/00cf30429597aed2551ebe8d07b14be276b41249/neurovault/apps/statmaps/views.py#L952

erramuzpe:
@chrisfilo Yes, hehehe that's the reason :) 

I'm having some errors in CircleCI since I merged with master yesterday. I saw some of your accepted PRs in the last days also having them, so I understand that you are OK with them. :+1: 

The error is in `test_authentication_allow`

Do you like the actual benchmark is now OK or something more is needed? I can not get rid of the "complexity" check fail from codacy...

erramuzpe:
@chrisfilo  FYI

[Results from queries](https://gist.github.com/erramuzpe/a1fc301e7466ffef85f9e3b5a41c3d2f)

Results are not "complete". It gives just the ones that are closer, but not the rest. I will try to search on this. This is just a toy-bench, not real data, not real results, not tested if results are accurate or just random (not real nothing... haha). But it is useful to give an idea of the scale and magnitudes we are talking about. I am super-impressed by the bench results (and Nearpy was one of the worst ANN libraries in the global bench...).

More info: https://erramuzpe.github.io/gsoc2016/2016/06/10/Testing-A-N-N-libraries.html

Benchmark:
nearpy_bench

erramuzpe:
@chrisfilo I got rid of the circleCI error. I can not figure out how to improve the quality to get the codacy one passed

erramuzpe:
@chrisfilo I am exploring dimensionality reduction for the data to go in the Engine (the vectors from reduced_dimensionality) and I tried as experiment to reduce the maps to 16mm voxels instead of 4mm ones I was using for the bench. Results are incredible good. 

For 940 images:

|  | 4x4x4mm  | 16x16x16  | 
| ------------- |:-------------:|:-------------:|
| Size (npy) | 214.7 MB | 3.4MB | 
|Vector dim | 28549 | 450  | 
|Indexing (mean) Nearpy| 0.025s | 0.0006s (*really*) |
|Query (mean) Nearpy| 0.0085s | ~0.0025s (Cosine) / ~0.005s (Euclidean) ?¿ | 

And not only that: 
Random query: Queried 1710 results -> [1710](http://neurovault.org/images/1710/), [533](http://neurovault.org/images/533/), [19128](http://neurovault.org/images/19128/), [184](http://neurovault.org/images/184/) .... (up to 100 or whatever number we want) 
They look quite "similar". 

Time taken to build the whole Engine for 940 images: 0.510115 seconds O.o :dancer::dancer:

chrisfilo:
You can ignore the Codacy score for now.

Impressive speed, but we cannot look at it without taking accuracy into consideration...

chrisfilo:
BTW did you try doing PCA and creating reduced representation using projections to previously estimated components?

erramuzpe:
I was trying to do it today (again), but I can not design a way of doing it, since I was thinking: 
If I have a projection of an image in a lower space (PCA), how can I do this for all the images and maintain the same lower space (because I want the images to be comparable)? The new lower space depends on the characteristics of the data, doesn't it?

chrisfilo:
You estimate the components once using a big dataset and use it later even for new images that were not included in the original PCA estimation. Does this make sense?

erramuzpe:
I think so, I was thinking in a single image approach. But what you say is like building a common space from a training dataset and this common space driving the dimensionality reduction even if it is not the optimal for the new image, but it is for the whole dataset. Is it something like this?

chrisfilo:
yes - this should work as long as PCA is perform on a big and
representative dataset

erramuzpe:
I made the PCA approach having 500 images for the fitting (so we are having 500 sized vectors to insert in the Engine). The explained variance ratio is 93% for the first component. This is also very fast (there is the need of computing the pca, but this is also superfast and has to be calculated just once for each image). 

To compare with yesterday's reduction, same query:

Queried 1710, results -> [1710](http://neurovault.org/images/1710/), [2744](http://neurovault.org/images/2744/), [533](http://neurovault.org/images/533/) (this was 2nd with yesterday's approach, which was totally different) , [113](http://neurovault.org/images/113/) ...

My impression is that this approach depends heavily in the projection orientation. I will make an extensive comparison with the actual implementation's results as soon as the PR for the find_similar results is accepted. 
- What kind of comparison should I perform? Is it OK to see the ratio of intersection looking to the first 100 results of each query or also the order matters?

chrisfilo:
when you said
> there is the need of computing the pca, but this is also superfast and has to be calculated just once for each image

did you mean calculating PCA or projecting a map to a space of the previously computed PCA?

> My impression is that this approach depends heavily in the projection orientation.

PCA should be rotation invariant so this does not matter.

The first component explains a lot of variance. Could you visualize it so we could see what is it capturing? BTW we should use more than one map...

For assesing accuracy have a look at: https://www.wikiwand.com/en/Discounted_cumulative_gain

erramuzpe:
> did you mean calculating PCA or projecting a map to a space of the previously computed PCA?

oh, yes, the projection.
> Could you visualize it so we could see what is it capturing?

ok
> we should use more than one map

what do you mean by this?
This is what I'm doing, I have a feature matrix of 940 subjects and each vector has length 28549
```
#features must be n_samples*n_features =  940 * 28549
number_of_samples = 500 #fitting maps -> this is what you mean by more than one?
pca = PCA()
pca.fit(features[:number_of_samples, :])
#print(pca.explained_variance_ratio_[:4])
example = pca.transform(features[501,:]).T # 1 example map projection, this is what I'm giving to the Engine
```

chrisfilo:
Sorry - I meant to write more than one component - maybe about 20 to start
with? From the code snipped it seems you are using all components at the
moment.

erramuzpe:
I will take a more careful look, [pca.trasform](http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA.transform).

erramuzpe:
@chrisfilo As I see in the DCG function, there is the need of giving weights to the ordered values. Do I gave them the absolute value of the correlation as it comes from the actual implementation? Other possibilities is to give a 1 to the first result and normalize the rest, or give weights depending on the order (for example). Also, if the results don't appear in the first 100 results, do I gave them a weight=0?

erramuzpe:
First 2 components. The points forming the axes ?¿ is this noise?
pca_n20

chrisfilo:
Something funky is happening here. How do the weights of the first two components look visualized as a brain map?

For the distance I think using abs value or square value of pearson r (for the current implementation) sounds reasonable. We might consider using all images instead of just first the 100 to avoid the problem of non overlapping vectors.

erramuzpe:
> Something funky is happening here. How do the weights of the first two components look visualized as a brain map?

Absolutely, I think I'm doing something wrong...

neurovault/apps/statmaps/management/commands/benchmark.py
```diff
@@ -0,0 +1,85 @@
...
+                    image_name=file,
+                    ignore_file_warning=True)
+            index_table[i] = t.interval
+            np.save(os.path.join(app_path, 'results_index'+ str(datetime.datetime.utcnow())[:10]), index_table)
+
+            factory = RequestFactory()
+            request = factory.get('/images/' + str(i) + '/find_similar')
+            t = Timer()
+            with t:
+                _dummy = find_similar(request, image.pk)
```
chrisfilo: Why not `get_similar_images`?

neurovault/apps/statmaps/management/commands/benchmark.py
```diff
@@ -0,0 +1,85 @@
...
+            query_table[i] = t.interval
+            np.save(os.path.join(app_path, 'results_query' + str(datetime.datetime.utcnow())[:10]), query_table)
```
chrisfilo: Why not CSV?

PR #519, opened 2016-06-05
Gene expression decoding

chrisfilo:
This is a port of [alleninf](https://github.com/chrisfilo/alleninf). Major differences from previous implementation:

- We use `reduced_representation` of the decoded map. It's a downsampled (4x4x4mm) and limited to MNI brain mask version of the original map.
- Gene expression values  from wells that point to the same voxel of the reduced representation are averaged. This is done independently for each donor brain.
- For each gene one probe is selected based on mean maximum expression (following advice from http://www.ncbi.nlm.nih.gov/pmc/articles/PMC3166942/)
- All genes are evaluated. To account for the number of tests FDR correction is applied.
- Percentage of variance explained (mean and std) is reported.

The results have been validated using a set of PET, SPECT, and myelin maps and their corresponding genes. Same correlation direction and match vs. mismatch as reported in http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0148744 are found. Effect sized are substantially smaller, because MENGA averages values over 15 brain regions before calculating the correlation. We decided not to do such averaging to avoid problems with interpreting such [ecological correlations](https://www.wikiwand.com/en/Ecological_correlation).

In addition I tested 4 more serotonin maps from another PET group. Three of them (WAY, HTT, HTR1A, HTR2A) showed expected results. One HTR1B did not show a strong relationship. However this gene has been excluded in some other AHBA analyses (see http://www.ncbi.nlm.nih.gov/pubmed/26068849). I suspect probes for this gene are not reliable.

All of the validation tests has been implemented as unit tests.

This is how the results view looks like:
genes

I also added a dropdown button to group all of the analyses one can do with a map:
dropdown


tyarkoni:
Sorry I'm only getting to this now. FWIW, I think it's a really bad idea to provide t and p values (and probably r^2 as well). We have no idea what the true degrees of freedom are, so any calculation is sort of a wild guess. But using the nominal DFs (i.e., number of voxels) to calculate statistics is almost certainly much too optimistic (assuming that's what you did here). In our paper, we explicitly argued _against_ reporting or using such values. The worry is that the numbers don't actually mean anything, but people are likely to draw all kinds of unwarranted conclusions. Basically, this feels like providing a soft license to go on a fishing expedition on an epic scale.

tyarkoni:
At bare minimum, we should provide a prominent disclaimer above the table that says something to the effect of "THESE ASSOCIATIONS SHOULD BE INTERPRETED WITH GREAT CAUTION AS THE TRUE DEGREES OF FREEDOM ARE UNKNOWN AND THE LIKELIHOOD OF A TRUE CAUSAL ASSOCIATION BEING PRESENT IS EXTREMELY LOW."

tyarkoni:
Ah, I just noticed your test is done over donor brains rather than over voxels. That addresses the point about the nominal DFs--sorry, I should have read the details first. That said, I still think we should put a prominent disclaimer at the top of the page saying the proportion of true causal associations is likely to be extremely low.

tyarkoni:
Also, I think it would be a good idea to link to the corresponding gene pages on Neurosynth and the ABA. E.g.,

http://human.brain-map.org/microarray/search/show?exact_match=false&search_term=CYP2D6&search_type=gene
http://neurosynth.org/genes/CYP2D6/

I think it's important for people to have an intuitive understanding of how much spatial uncertainty there is about these expression maps before drawing conclusions.

tyarkoni:
Presentation/interface looks great though!

Issue #207, opened 2015-04-22
Add comparison to the API


vsoch:
If we eventually want comparison to drive meta analysis, a user should be able to find the most similar images programmatically (via the API)

erramuzpe:
@chrisfilo I am trying to solve this and I made some progress. I can retrieve the info for a specific comparison information using the API, but manually. I mean, I statically select an image, and gave the info about comparisons through the API. 

```
class ComparisonViewSet(viewsets.ModelViewSet):

    max_results = 100
    queryset = get_existing_comparisons(PK=3).extra(select={
          "abs_score": "abs(similarity_score)"}).
          order_by("-abs_score")[0:max_results]  # NOTE PK=3 -> this is the "static"
    serializer_class = ComparisonSerializer
    permission_classes = (ObjectOnlyPolymorphicPermissions,)
```
```
class ComparisonSerializer(serializers.Serializer):
    id = serializers.ReadOnlyField(source='image2.pk')
    abs_score = serializers.ReadOnlyField()
```

with this, I get the paired information in `http://127.0.0.1/api/comparison` , great. BUT I encountered some troubles when trying to extend this for `http://127.0.0.1/api/comparison/{pk}` . There are 2 kind of problems.

- The first one has relation with the way of retrieving the information and how the REST API works. As I have understood, the API works with a general query; say `/api/comparison` and then, you can filter it with using `/api/comparison/{pk}` using the `@detail_route` method or a retrieve function, right? This is prepared for a Model/QuerySet based architecture. The problem here is that we don't need the first part, just the part where the` {pk}` is used. I can not make it work in this way as is asking me a first query.

- The second one is conceptual; so, even if I manage to solve the first problem for the actual implementation, the aim is to substitute the Comparison model for an Engine that can be queried, but we are not going to get responses as QueryObjects/Models and the integration in the API looks difficult to adapt. My question is: is there any way of building a ViewSet for the API without using the Model architecture and getting rid of  `/api/comparison` ? 

Sorry for the mess, I hope all this makes sense to you!

chrisfilo:
You are right about the lack of need for the Comparison model in one of the implementations. I would recommend to create JSON replies without using the Django REST Framework (see how atlas and gene expression APIs are implemented). It would make replacing this implementation easier later on.

erramuzpe:
Thanks!

PR #529, opened 2016-06-22
Find similar results as json

erramuzpe:
/images/{pk]/find_similar/json
returns a pandas dataframe as json as in gene_expression_json -> {"data":[[1.0,3.0],[-0.192888910181686,16.0],[0.147546693160353,13.0],[0.083965640541661,14.0],[-0.0595162102504215,17.0],[-0.0354389236538608,15.0]],"columns":["abs_value","image_id"]}

This PR solves #207

### DIFF
--- a/neurovault/api/serializers.py
+++ b/neurovault/api/serializers.py
@@ -13,8 +13,7 @@ from neurovault.apps.statmaps.forms import (handle_update_ttl_urls,
 from neurovault.apps.statmaps.models import (Atlas, Collection, Image,
                                              NIDMResults,
                                              NIDMResultStatisticMap,
-                                             StatisticMap, BaseCollectionItem,
-                                             Comparison)
+                                             StatisticMap, BaseCollectionItem)
 
 
 class HyperlinkedFileField(serializers.FileField):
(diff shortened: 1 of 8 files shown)


erramuzpe:
https://github.com/NeuroVault/NeuroVault/blob/951dcac11df090d610f35f838465f2f1e7113c36/neurovault/apps/statmaps/views.py#L1003-L1007

I can delete the lines straightforward in this same PR if you want

chrisfilo:
Couple of things:
1. Could you refactor the `find_similar` view to use the new util function `calculate_image_similarity`?
2. Could you rename the `calculate_image_similarity` to `get_similar_images` or something along those lines (since "calculate" is misleading in the current implementation).
3. Could you add a test or two?

erramuzpe:
1.- Yes, but how do I do; 1 return for the json case and another return for the view case?
2.- Ok! sorry
3.- I will try, first time doing it :)

chrisfilo:
As far as I see `calculate_image_similarity` currently returns a pandas dataframe not a JSON. So after some refactoring it could be used inside `find-similar` view (note that I did not suggest replacing `find_similar`)

erramuzpe:
There is a problem with "tags", I will solve it tomorrow!

erramuzpe:
Solved the problem with tags. Let's see how the tests work

erramuzpe:
@chrisfilo check if you like the test . In principle, this and the benchmark are ready to merge.

PR #533, opened 2016-07-04
[WIP - URGENT] update and refactor find_similar() 

erramuzpe:
This is needed, since we don't want to limit the comparison between metrics to the first 100 results. 
- 100 results limit for the find_similar view (html)
- no limit for json response


erramuzpe:
@chrisfilo please, whenever you have time, take a look into this and upload to production. I will need it to perform the comparisons. Thanks in advance!

chrisfilo:
Why do you need those changes to be deployed in production to perform the accuracy benchmark? It should work equally well offline on the example test

neurovault/apps/statmaps/views.py
```diff
@@ -1002,7 +1002,7 @@ def find_similar_json(request, pk, collection_cid=None):
 
     image1 = get_image(pk, None, request)
     pk = int(pk)
-    max_results = 100
+    max_results = Image.objects.all().count()  # image number will always be higher than comparison number
```
chrisfilo: This will result in the find_similar view showing all images instead of the first 100. Page will be huge and will load slow.

erramuzpe:
Yes, that's true, but I've limited computational power to perform the calculations of Comparisons for 940 images (which is the amount of subjects I've chosen for the bench, no specific reasons). I was thinking in using production since there it is already calculated.

neurovault/apps/statmaps/views.py
erramuzpe: Just the json, not the html. We said the comparison should be over the whole set and not just the first 100, but we can limit to the first 100, no probs.

chrisfilo:
I would prefer to keep it separate from the production server. You must've
already calculated the comparisons for when you did the benchmark (at least
for a subset of maps). Maybe you can reduce the benchmark set to get more
reasonable run times?

erramuzpe:
I've been playing with clearDB() all this time hehe but anyway, ok, I'll reduce the set, run the comps and try the comp-bench on it. I'm out of the lab until Monday, but I'll set up everything for a limited set of images and once in the lab (with a better machine than my laptop), I'll increase the set a bit. :+1: 

So, if we are not using this PR, shall I close it?

PR #534, opened 2016-07-21
Cognitive decoding using sets of topic maps

chrisfilo:
This PR introduced a new way of doing cognitive decoding - using a set of topic maps. Each topic map has a dictionary of term/weight pairs assigned to it. During the decoding the map in question is projected to the set of topic maps and projection weights are combined with the term weights. Currently this approach can be applied to topic maps from Yeo et al. 2015 and Rubin et al. (submitted). New sets of topic maps can be easily added, by creating a new collection with the right attributes.

New analysis drop down menu.
2016-07-21 16_33_25-neurovault_ zstat old young

Page for choosing the decoding method. I was testing it using Yeo et al. 2015 maps, but other sets can be easily added. In production we will feature the Rubin et al. maps.
2016-07-21 16_33_53-atlases and parcellations

Results of the decoding
2016-07-21 16_35_06-cognitive decoding_ image name

Open questions/ TODO:
- Tests. What is the best way of testing this feature? Do we have a map with expected decoding results?
- Term links. It would be nice that terms were clickable and leading to more information (Cognitive Atlas or neurosynth)

@tyarkoni and @yeoyeo02 I would love your feedback.


chrisfilo:
```
$ docker-compose.exe up -d
$ docker-compose exec worker bash
# python manage.py fetch_yeo
```

PR #536, opened 2016-08-10
[WIP] Nearpy migration

erramuzpe:
## Migration PR for integration of NearPy as new Engine for find_similar

I have prepared a [document](https://docs.google.com/document/d/1IZXB52lUkAfPS0OVsxC1chAvnJnogJDduua33GTlXAc/edit?usp=sharing) explaining some major aspects involving this PR. 
### Instructions

While the PR is in [WIP], a good way of testing is to download the Docker image of Neurovault and perform the following instructions (also valid for Production) after running the command _populate_db_ through _./manage.py_

1.- Run the migration # 71 (deletes Comparison and Similarity _Models_)
2.- Understand the implications of dimensionality reduction of _reduced_representation_ . 
3.- Run the desired dimensionality reduction: _./manage.py change_resample_dim VOXEL_SIZE_
(I would recommend a VOXEL_SIZE of 16, according to performance and accuracy benchmarks)
4.- Run the desired Engine Building command:
-  _./manage.py rebuild_engine_ for Random Binary Projection hashes (recommended)
-  _./manage.py rebuild_engine_pca_ for PCA Binary Projection hashes (atm only works for 16mm voxel size _reduced_representation_ )

In both cases, the main parameters involved in the construction of the Engine are pre-set, but can be changed. 
### TO DO
- Wait for new version of NearPy and change the deletion function.
- Find a way of getting rid of load/saving the Engine through pickle if this solution does not fit us.


erramuzpe:
_test_counter_ and _test_thresholded_comparisons_ must be updated

erramuzpe:
I did not know that calculate_gene_expression_similarity() was also using the reduced_representation. This rises an error when using the reduced_representation (I'm resampling it to an even lower dimensionality, for the Engine). What would be a solution? to create a new field for another reduced_representation specific for the Engine? This is something to discuss.

erramuzpe:
I was working in failure of tests the whole day long and finally realized that `tearDown()` does not run if the test arises an ERROR, leading to errors in following tests. Due to this unnoticed behaviour I've changing some variables and "cleaning" tests. Even if the changes I made won't alter the tests, I will undo the mess soon.

chrisfilo:
A few things:

- We need to keep the 4x4x4 reduced representation files for the urpose of the gene expression analysis (at least for now). If you think it's necessary you might create (in memory) another resampling just for the purpose of indexing.

- It would be great if you could get the tests to pass.

- I am a bit concerned to merge this while the engine is kept in memory - I'm afraid it will not be sustainable (we already often run out of RAM on the production server even without this feature on). Having it working with redis would be much better.

- Similarly the question of rebuilding the hashes when the database grows (to adjust bucket sizes) remains unsolved. Do you have some ideas for a strategy to make this happen?

erramuzpe:
> We need to keep the 4x4x4 reduced representation files for the

Ok, we don't actually need the reduced representation in memory for the Engine, as the vectors are saved in the Engine; but I think it is a good idea to keep them, as it will make faster a rebuilt of the Engine. I will kept the 4x4x4 in memory as it was before and create a new reduced representation for the Engine. 

> It would be great if you could get the tests to pass.

This will be solved with the previous point :+1: 

> I am a bit concerned to merge this while the engine is kept in memory. Having it working with redis would be much better.

At the moment, the Engine is saved after every indexing operation trough pickle and loaded before every query, so, it is persistent (and fast). I am not sure if this is a desirable behaviour and I also think that Redis would be a better solution. But, Redis delete_vector() method is not implemented yet (I will, but even if I make it this week, it will take time to be in the new release of NearPy and testing... it will take a bit of time). Also, it is needed to be tested if that implementation fits our needs, since Redis does not save the Engine itself, but the hashes (so the Engine object has to be created before every index and query), and I don't know if this would be faster than saving and loading the pickle (this has to be testes, I will). 

> Similarly the question of rebuilding the hashes when the database grows (to adjust bucket sizes) remains unsolved. Do you have some ideas for a strategy to make this happen?

This is an operation that has to be performed very few times a year, and I though it could be done manually (but it can be programmed as well). What I was thinking is to decide a good mean images/bucket ratio and rebuild the engine every time it reaches a threshold. 

I decided to fix 7 bits hash-codes at the moment, which makes 128 possible buckets/hash and gives a mean images/bucket ratio of 40 for ~5000 images (actual size of NeuroVault's comparable images). To keep that ratio, we would have to rebuild the Engine with one bit more per hash-code every time the database doubles. Also, taking a closer look visually to the results, I might have been too conservative, I thing that a higher mean images/bucket ratio could avoid some false negatives. I will check this in more detail.

PR #538, opened 2016-08-13
Fix a server error on an empty string metadata value

### LINKED ISSUE #526: server error with edit metadata table
To reproduce, go to editmetadata view:

http://neurovault.org/collections/1029/editmetadata

image

Enter 34 and then 48 under "No. of subjects" and click save.

Likely it's an issue with some value already in the data that is trying to be saved.

burnash:
This PR fixes #526 and adds a test for the case.


chrisfilo:
Deployed.

Issue #495, opened 2016-05-02
"Download all images" button


chrisfilo:
Similar to https://github.com/NeuroVault/NeuroVault/issues/228, but simpler.

chrisfilo:
This could provide a streaming solution that would not require making temporary files: https://github.com/allanlei/python-zipstream

Requested by @GaelVaroquaux

GaelVaroquaux:
Cc @jeromedockes

PR #551, opened 2016-09-18
Added download collection feature.

erramuzpe:
Added a button in the Collection view that allows to download the whole Collection in a zip file on the fly.

### DIFF
--- a/neurovault/apps/statmaps/tests/test_collections.py
+++ b/neurovault/apps/statmaps/tests/test_collections.py
@@ -10,7 +10,7 @@ from uuid import uuid4
 
 from neurovault.apps.statmaps.models import Collection, User, Image, Atlas
 from neurovault.apps.statmaps.utils import detect_4D, split_4D_to_3D
-from neurovault.apps.statmaps.views import delete_collection
+from neurovault.apps.statmaps.views import delete_collection, download_collection
 from neurovault.settings import PRIVATE_MEDIA_ROOT
 from .utils import clearDB, save_statmap_form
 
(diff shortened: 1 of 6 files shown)


neurovault/apps/statmaps/views.py
```diff
@@ -1196,3 +1197,40 @@ class MyCollectionsJson(PublicCollectionsJson):
...
     def get_initial_queryset(self):
         return get_objects_for_user(self.request.user, 'statmaps.change_collection')
+
+
+def download_collection(request, cid):
+    # From https://stackoverflow.com/questions/67454/serving-dynamically-generated-zip-archives-in-django
+
+    # Files (local path) to put in the .zip
+    collection = get_collection(cid, request)
+    filenames = [ img.file.path for img in collection.basecollectionitem_set.all()]
```
erramuzpe: Depending on the configuration of production, this might be 

`filenames = [ img.file.url for img in collection.basecollectionitem_set.all()]` , not sure, but it works as it is now for me.

erramuzpe:
This PR is to solve #495

chrisfilo:
The current implementation, even though it would probably work, involves storing the complete zip file in memory. We have some big collections in Neurovault (2-3GB) which would become problematic. I would recommend using [zipstream](https://github.com/allanlei/python-zipstream) instead as initially suggested in #495.

Also please add at least one test.

erramuzpe:
@chrisfilo I added a test that calls the response, but I couldn't manage to actually make the file to download. What I am checking is that the response is not empty (I've checked that calling the method getvalue() of the response, you can actually get the zip file). The docker container has to be rebuilt, since there is a new module installed. 
Let's wait for the tests to be completed, I got an strange error in the previous push...

erramuzpe:
@chrisfilo tests passed :)

neurovault/apps/statmaps/tests/test_collections.py
```diff
@@ -320,3 +320,31 @@ def test_incorrect_value_in_fixed_foreign_field(self):
...
+        self.user.delete()
+
+    def testDownloadCollection(self):
+        self.client.login(username=self.user)
+        pk1 = self.Collection1.pk
+        request = self.factory.get('/collections/%s/download' %pk1)
+        request.user = self.user
+        response = download_collection(request, str(pk1))
+
+        self.assertTrue(len(response.getvalue())) # If there is something in the response, the file was generated.
```
chrisfilo: could you check if the status code in the response is 200?

neurovault/apps/statmaps/urls.py
```diff
@@ -49,6 +49,9 @@
     url(r'^collections/(?P<cid>\d+|[A-Z]{8})/delete$',
         delete_collection,
         name='delete_collection'),
+    url(r'^collections/(?P<cid>\d+|[A-Z]{8})/download$',
```
chrisfilo: I wonder if it would make sense to change this to reflect the fact that the response is a zip file.  For example replace "download" with `"%s.zip" % collection.name`?

neurovault/apps/statmaps/tests/test_collections.py
erramuzpe: Yes!

erramuzpe: I think I can do the API link too

neurovault/apps/statmaps/urls.py
erramuzpe: Well, I think it could be done, but that could be problematic for certain cases. eg. Somebody that wants fastly to download a Collection, if we add the name of the collection, it is quite large in some cases... Not sure about it. Just let me know :)

chrisfilo: Ok let's leave it as it is for now.

Issue #555, opened 2016-09-22
Download Collections bug - NIDM Results


erramuzpe:
Server Error when downloading [NIDM Results](http://neurovault.org/collections/1404/downloads)

PR #556, opened 2016-09-22
Solves #555, NIDM Results download feature. Added tests

erramuzpe:
Solves #555, it was complaining because nidm is not an statistical map set and could not access to the files. As it is now, when downloading, it takes just the nidm_result.zip (which includes the statistical map and the ttl). Also I added a test for this.


chrisfilo:
Thanks for sending this. The problem with the current implementation is that it does not work well with collections that include both NIDMREsults objects and normal statistic maps.

I would suggest doing the following: when looping over base_collection_items:

- if the object is a subclass of Image, but not NIDMStatisticMap add it to the zip via .file field
- if the object is a NIDMResults class add it to the zip via .zip_file field

erramuzpe:
@chrisfilo Hopefully solved. Not best idiomatic python, but it was also a long sentence... 
So, now all Image instances but NIDMStatisticMap are being added.

neurovault/apps/statmaps/views.py
```diff
@@ -1202,16 +1202,18 @@ def get_initial_queryset(self):
...
-    from django.http import StreamingHttpResponse
-
     # Files (local path) to put in the .zip
     collection = get_collection(cid, request)
-    filenames = [img.file.path for img in collection.basecollectionitem_set.all()]
+
+    filenames =[img.zip_file.path for img in collection.basecollectionitem_set.all()
+                if isinstance(img, NIDMResults)] + \
+               [img.file.path for img in collection.basecollectionitem_set.all()
+                if (isinstance(img, Image) and not isinstance(img, NIDMResultStatisticMap))]
```
chrisfilo: You can incorporate this into querysets using django polymorphic https://django-polymorphic.readthedocs.io/en/stable/advanced.html It should yield higher performance.

neurovault/apps/statmaps/views.py
```diff
@@ -1202,16 +1202,17 @@ def get_initial_queryset(self):
...
 def download_collection(request, cid):
-    from django.http import StreamingHttpResponse
-
     # Files (local path) to put in the .zip
     collection = get_collection(cid, request)
-    filenames = [img.file.path for img in collection.basecollectionitem_set.all()]
+
+    filenames = [img.zip_file.path for img in collection.basecollectionitem_set.instance_of(NIDMResults)] + \
+                [img.file.path for img in
+                 (collection.basecollectionitem_set.instance_of(Image).exclude().instance_of(NIDMResultStatisticMap))]
```
chrisfilo: The empty `.exclude()` does not do anything. You should use `not_instance_of(NIDMResultStatisticMap)` instead. Coudl you add a test that downloads a collection with NIDMResults as well as normal StatisticMaps (not coming from a NIDM Results object) and makes sure all of the expected files are present in the downloaded zip?

erramuzpe: It was working in that way in my local, not sure why... Anyway, I corrected it, thanks!
About the test, I was trying many ways of doing it, but I couldn't achieve it, I need help. 

So, I have this `response`, which has a field called `response.streaming_content`, an iterable with the data. I'm reading it as:

```
import zipfile
with io.StringIO() as fl:
    for chunk in response.streaming_content:
        fl.write(chunk)
    zf = zipfile.ZipFile(fl)
```
and I'm getting a `BadZipfile: File is not a zip file` error. I've tried many different solutions, any idea?

erramuzpe: Solved, give me 10 mins :)

neurovault/apps/statmaps/tests/test_nidm.py
```diff
@@ -106,62 +109,105 @@ def tearDown(self):
...
+    #         nidm = form.save()
+    #
+    #         self.assertEquals(len(nidm.nidmresultstatisticmap_set.all()),
+    #                           info['num_statmaps'])
+    #
+    #         map_type = info['output_row']['type'][0]
+    #         map_img = nidm.nidmresultstatisticmap_set.filter(
+    #             map_type=map_type).first()
+    #
+    #         self.assertEquals(map_img.name, info['output_row']['name'])
```
chrisfilo: Could you remove this block or describe why is it commented out?

erramuzpe: :fearful: Sorry about that, I commented that block and forgot to uncomment.

PR #558, opened 2016-10-04
Switch to Anaconda and Intel (MKL) optimized libraries

chrisfilo:


chrisfilo:
Switching to optimized intel binaries does not seem to show speed ups - at least not in time it takes to run tests on CircleCI.

PR #566, opened 2016-11-12
Order the queryset results by the collection item id

### LINKED ISSUE #562: Duplicate images in paginated API endpoint
It looks like `/collection/<id>/images/` endpoint sometimes returns same images on different pages.

In instance, an image with id [27630](neurovault.org/images/27630/) appears in 
http://neurovault.org/api/collections/1806/images/  (aka http://neurovault.org/api/collections/1806/images/?limit=100&offset=0) _and_ also in the results on the `next` page: http://neurovault.org/api/collections/1806/images/?limit=100&offset=100

I guess the inconsistency is caused by the missing ordering clause in the source queryset. I can fix that but I need to confirm what model field to use for ordering. I suggest `add_date` since it's a common field for all models derived from `BaseCollectionItem`.

burnash:
fixes #562

### DIFF
--- a/neurovault/api/views.py
+++ b/neurovault/api/views.py
@@ -300,7 +300,7 @@ class CollectionViewSet(viewsets.ModelViewSet):
 
     def _get_paginated_results(self, obj_class, pk, request, serializer):
         collection = get_collection(pk, request, mode='api')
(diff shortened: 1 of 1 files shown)


burnash:
@chrisfilo CI strangely fails for this simple change. Although all tests pass locally.

PR #569, opened 2016-11-22
Downloaded files keep the full name for the zip file.

erramuzpe:
For Collections with names `test_collection` the generated zipfile name is `test_collection.zip`. 
For Collections with names `test collection` the generated zipfile name is `test` due to the space between words. 

Small change to correct this issue, replacing spaces for underscores.

### DIFF
--- a/neurovault/apps/statmaps/views.py
+++ b/neurovault/apps/statmaps/views.py
@@ -1224,7 +1224,7 @@ def download_collection(request, cid):
         zf.write(fpath, zip_path)
 
     response = StreamingHttpResponse(zf, content_type='application/zip')
-    response['Content-Disposition'] = 'attachment; filename=%s' % zip_filename
+    response['Content-Disposition'] = 'attachment; filename=%s' % zip_filename.replace (" ", "_")
     return response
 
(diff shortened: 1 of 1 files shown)


erramuzpe:
Another solution: Django has a 'slugify' function which does something similar:

```
from django.template.defaultfilters import slugify
slugify("This should be connected")
this-should-be-connected
```

neurovault/apps/statmaps/views.py
```diff
@@ -1224,7 +1224,7 @@ def download_collection(request, cid):
         zf.write(fpath, zip_path)
 
     response = StreamingHttpResponse(zf, content_type='application/zip')
-    response['Content-Disposition'] = 'attachment; filename=%s' % zip_filename
+    response['Content-Disposition'] = 'attachment; filename=%s' % zip_filename.replace (" ", "_")
```
chrisfilo: urlencode the string?

erramuzpe: `urllib.quote_plus(zip_filename)` gives `+` instead of spaces and
`urllib.pathname2url(zip_filename)` gives `%20` instead of spaces.

I will use the first one ¯\_(ツ)_/¯

PR #586, opened 2017-03-08
tweak gene expression to link to Neurosynth

tyarkoni:
Here's a first pass at this. Works fine but may need some tweaking as it's kind of ugly. Feel free to modify, or suggest a better way to present the info. Maybe we should add extra table columns for NCBI and NS links? Or shorten Neurosynth to 'NS'?

### DIFF
--- a/neurovault/apps/statmaps/templates/statmaps/gene_expression.html.haml
+++ b/neurovault/apps/statmaps/templates/statmaps/gene_expression.html.haml
@@ -19,7 +19,7 @@
         			"columnDefs": [
 										{
 											"render": function ( data, type, row ) {
-												return "<a href='http://www.ncbi.nlm.nih.gov/gene/" + row[1] +"'>" + data + "</a>";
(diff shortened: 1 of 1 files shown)


neurovault/apps/statmaps/templates/statmaps/gene_expression.html.haml
```diff
@@ -19,7 +19,7 @@
         			"columnDefs": [
 										{
 											"render": function ( data, type, row ) {
-												return "<a href='http://www.ncbi.nlm.nih.gov/gene/" + row[1] +"'>" + data + "</a>";
+												return data + " <a href='http://www.ncbi.nlm.nih.gov/gene/" + row[1] +"'>[NCBI]</a> <a href='http://neurosynth.org/genes/" + data +"'>[Neurosynth]</a>";
```
chrisfilo: Doesn't this prevent gene names to be displayed?

tyarkoni: No, it's just displayed before the links. So it looks like:

PPTC7 [[NCBI](http://www.ncbi.nlm.nih.gov/gene/160760)] [[Neurosynth](http://neurosynth.org/genes/PPTC7)]

chrisfilo:
oh ok

PR #590, opened 2017-04-16
Add a `private` property to the collections api response

burnash:
This PR adds `private` property to collection objects in the API response. 

It went missing in 371fa8539a6aecbbe784da3f703f6af96487b775. Is the reason for excluding it still relevant?  Now as we have `my_collections` endpoint, `private` could be quite useful.

If it's OK to bring it back, I'll add some tests to the PR.


chrisfilo:
Don't see a reason why to exclude it.

burnash:
Great, let me add some tests

burnash:
Finished with tests.

PR #592, opened 2017-06-05
Removing reference to requirements.txt in README

vsoch:
I noticed the requirements.txt is no longer present, but it's referenced in the README. This is a small tweak to adjust that.


chrisfilo:
Thanks!

PR #593, opened 2017-06-06
Fix import of NIDM-Results pack with F-tests

cmaumet:
This PR fixes an issue identified by @AlexBowring.

**Description of the problem**: When uploading a NIDM-Results pack containing an F-test but no T-test (e.g. [this one](https://drive.google.com/file/d/0B5rWMFQteK5edHlaZUM1TTVIbzg/view?usp=sharing)), the upload fails and NeuroVault returns "Invalid NIDM-Results file".

**Fix**: When loading the statistic maps available in a NIDM-Results pack, NeuroVault is looking for contrast maps to retrieve the corresponding contrast name. But contrast maps are only defined for T-tests (cf. [Contrast estimation in NIDM-Results](http://nidm.nidash.org/specs/nidm-results_130.html#h-contrast-estimation)). This PR updates the query to get the contrast name directly from the statistic maps. This solves the issue (tested on 1 example only) and should insure that the correct contrast name is associated with each statistic map.

Please let us know what you think of this. Thanks!

### DIFF
--- a/neurovault/apps/statmaps/nidm_results.py
+++ b/neurovault/apps/statmaps/nidm_results.py
@@ -63,18 +63,13 @@ class NIDMUpload:
         prefix prov: <http://www.w3.org/ns/prov#>
         prefix nidm: <http://purl.org/nidash/nidm#>
 
-        prefix contrast_estimation: <http://purl.org/nidash/nidm#NIDM_0000001>
-        prefix contrast_map: <http://purl.org/nidash/nidm#NIDM_0000002>
         prefix contrast_name: <http://purl.org/nidash/nidm#NIDM_0000085>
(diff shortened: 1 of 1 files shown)


chrisfilo:
Thanks!

cmaumet:
Thanks for merging @chrisfilo! Do you have an idea of when this fix will be available in production?

chrisfilo:
It should be now

cmaumet:
Great! Thanks! @AlexBowring: Can you check this solves your issue?

AlexBowring:
The issue is fixed, I just uploaded our NIDM-Results pack here: 
[http://neurovault.org/collections/2488/](http://neurovault.org/collections/2488/)

Thank you!

Issue #299, opened 2015-07-14
comparison metrics are not updated when a map gets updated


chrisfilo:

vsoch:
I don't know if this is relevant, but when I was testing "update_or_create" for my CogPheno database, the function would sometimes generate an entirely new object. That could be be why we are getting the errors that comparison objects already exist, and then why they aren't updated when they should be.

chrisfilo:
Steps to reproduce

1. Add two collections, both of the should have DOI field filled.
2. Add one map to each collection
3. Use "Find similar" to record the similarity score between the maps
4. Look at the scatterplot at http://neurovault.org/images/compare/<id_a>/<id_b>
5. Edit one of the maps and replace the Nifti file with a brand new file
6. Similarity score should change (but it doesn't).

Best way to start addressing this issue is to write a test first.

shubhsingh594:
@chrisfilo , can i try my hand at this ?

chrisfilo:
Be my guest!

vsoch:
![image](https://cloud.githubusercontent.com/assets/814322/21598629/aa22ea84-d114-11e6-8e6f-9a04a21a5a95.png)

vsoch:
:musical_note: ! :)

PR #595, opened 2017-06-14
Issue 299

erramuzpe:
Test for #299 . The test works fine without changing anything in the code. I've been trying to reproduce the bug vainly. 

I though that `run_voxelwise_pearson_similarity.apply_async([image.pk])` had to be added to [this](https://github.com/NeuroVault/NeuroVault/blob/master/neurovault/apps/statmaps/models.py#L393-L395) clause, but seems like it is not needed. 
:thinking: 

BTW: I'm getting an error in the tests.
`test_adding_nidm:
ValueError: The NIDMResults could not be created because the data didn't validate.
`


erramuzpe:
When I said "trying to reproduce", I meant that I was uploading

erramuzpe:
oops, wrong button

PR #601, opened 2017-10-30
[WIP] Adding ability to specify target space

chrisfilo:


chrisfilo:
migration files seem to be missing

neurovault/apps/statmaps/models.py
```diff
@@ -452,6 +459,7 @@ class BaseStatisticMap(Image):
                     verbose_name="Map type",
                     max_length=200, null=False, blank=False, choices=MAP_TYPE_CHOICES)
 
+    subject_species = models.CharField( max_length=200, blank=True, null=True)
```
chrisfilo: Let's move `target_template_image` here as well.

neurovault/apps/statmaps/templates/statmaps/statisticmap_details.html.haml
```diff
@@ -161,9 +161,15 @@ a.disabled-link:hover {
                     Analysis
                     %span.caret
                 %ul.dropdown-menu
-                    %li
-                        %a{:href => "http://neurosynth.org/decode/?neurovault={{ api_cid }}"}
-                            %div.tooltip-wrapper{style: "display: inline-block", data-toggle: "tooltip", title: "Cognitive decoding using coordiante based data collected from thousands of papers."}
+                    - if neurosynth_compatible
```
chrisfilo: we should do the same for gene decoding and find similar

neurovault/apps/statmaps/models.py
```diff
@@ -26,6 +26,11 @@
 from neurovault.settings import PRIVATE_MEDIA_ROOT
 
 
+# I wanted to put this in utils, but it had to go here to avoid circular imports. :-/
+### I think it should probably be read from a template object? 
+def get_target_template_list():
+    return [('MNI152', 'Human'), ('NMT', 'Rhesus (macacca mulatta)'), ('Dorr_2008_average', 'Mouse (Dorr et al., 2008 space)')]
```
chrisfilo: `'MNI152NLin2009cAsym', 'Human (MNI152 NLin 2009c Asym)'`

chrisfilo:
Reducing template size
- cropping using nilearn
- reducing precision to int8
- downsampling

chrisfilo:
This functions should check if the image is in MNI and exit without doing anything if not:

https://github.com/NeuroVault/NeuroVault/blob/2cc22e24c64068e3ef129ef04f96a76401ab6591/neurovault/apps/statmaps/tasks.py#L152

This function should return an empty list if the image passed in the argument is not in MNI:
https://github.com/NeuroVault/NeuroVault/blob/2cc22e24c64068e3ef129ef04f96a76401ab6591/neurovault/apps/statmaps/utils.py#L516

neurovault/apps/statmaps/models.py
asfox: Is this another template? I'll add it, but you'll need to share the template with me. :-)

chrisfilo: nope - it's the more accurate description of the current template

Issue #607, opened 2018-01-16
Excursion set not displayed in NIDM-Results collections


cmaumet:
Hi,

@satra reported that the excursion sets are not displayed for the NIDM packs in this collection: https://neurovault.org/collections/1425/ 

For example, with https://neurovault.org/collections/1425/pain_01.nidm , we get a red message saying "There was a problem reading that file (ExcursionSet.nii.gz)".

I've tried to debug the issue but have not managed so far. I can reproduce the problem on a local dev instance of NeuroVault but not on a local dev instance of the nidmviewer.

@vsoch, @chrisfilo: would you have any recommendation of things I could look at to try finding a patch?

chrisfilo:
Might be worth looking into JS error console with Chorme Developer Tools.

Best,
Chris

cmaumet:
Thanks! You were both right! I've sent a PR to the nidmviewer: https://github.com/vsoch/nidmviewer/pull/19.

cmaumet:
(and also a small edit to NeuroVault: https://github.com/NeuroVault/NeuroVault/pull/608).

chrisfilo:
We currently have the following config:
```
server {

  listen                80;
  listen [::]:80;
  server_name           neurovault.org;
  location / {
    rewrite ^ https://$host$request_uri? permanent;
  }
}
```

Should I switch rewrite to 301?

satra:
@chrisfilo - it's the appropriate http redirect message.

regarding the current issue, i see an appropriate redirect with wget. therefore it's likely that whatever is in the code that's pulling the file is not able to follow. i'll take a look in a second.

```
$ wget http://neurovault.org/collections/NJXQPEEK/pain_01.nidm/ExcursionSet.nii.gz
--2018-01-22 21:35:05--  http://neurovault.org/collections/NJXQPEEK/pain_01.nidm/ExcursionSet.nii.gz
Resolving neurovault.org... 54.149.190.168
Connecting to neurovault.org|54.149.190.168|:80... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://neurovault.org/collections/NJXQPEEK/pain_01.nidm/ExcursionSet.nii.gz [following]
--2018-01-22 21:35:05--  https://neurovault.org/collections/NJXQPEEK/pain_01.nidm/ExcursionSet.nii.gz
Connecting to neurovault.org|54.149.190.168|:443... connected.
HTTP request sent, awaiting response... 200 OK
[...]

PR #608, opened 2018-01-17
Change 'http' into 'https' in domain name

cmaumet:
This is a very simple update which changes 'http' into 'https' into the `DOMAIN_NAME` variable. 

This will ensure that any NIDM pack uploaded will include https paths to the images (rather than http  then redirected to https).

*But* I don't know how this will impact the rest of NeuroVault so please feel free to ignore if you think that's a bad idea. (The updated nidmviewer should work regardless.)

### DIFF
--- a/neurovault/settings.py
+++ b/neurovault/settings.py
@@ -13,7 +13,7 @@ BASE_DIR = os.path.dirname(os.path.abspath(__file__))
 
 DEBUG = True
 
-DOMAIN_NAME = "http://neurovault.org"
+DOMAIN_NAME = "https://neurovault.org"
 
 TEMPLATE_DEBUG = DEBUG
 


chrisfilo:
Deployed!

cmaumet:
Great!! Thank you! 

This fix solves the problem of display (#607) for newly updated NIDM packs (cf. test collection at: https://neurovault.org/collections/HYZWIICI/pain_01.nidm). 

Now let's see if we can solve the issue for NIDM packs that are already on the system :) Discussion continues at https://github.com/vsoch/nidmviewer/pull/19 and https://github.com/NeuroVault/NeuroVault/issues/607...

vsoch:
+1 here, https everywhere is becoming less of recommendation, more of "mandatory or you lose all the cool things for your website!"  --> https://blog.mozilla.org/security/2018/01/15/secure-contexts-everywhere/

PR #614, opened 2018-03-01
Correct FAQ formatting

dankessler:
The last FAQ was added in #576, but it looks like the answer didn't make it into a `.well` class. This PR fixes that. 

It'd probably also be nice to make the suggested code stand out from the rest of the text, à la
>Use this command line FSL tool: `fslcpgeom MNIatlas image_with_weird_header`. (Many thanks to Leonie Lampe for the tip).

but I don't know enough about HAML to accomplish this (maybe `` `code` `` would be enough, but I couldn't quickly figure out how markdown-enabled HAML was out of the box)


chrisfilo:
Thanks!

Issue #630, opened 2018-06-30
Upgrade to Python 3


chrisgorgo:
NeuroVault is currently using Python 2 which is outdated and soon will become deprecated. We need to switch to Python 3.

The transition will require modification of Dockerfile and applying automatic code conversion to the codebase. Automatic tests should pick on any problems.

The transition might require upgrading/fixing couple of custom dependencies:
- nidmresults
- pybraincompare
- pycortex

I started working on this at https://github.com/chrisfilo/NeuroVault/tree/enh/python3 - this work could be picked up from there or started from scratch. I managed to convert all of the NeuroVault code, but got stuck dealing with external dependencies.

Transition to Python 3 needs to not come at a cost of loosing any functionality or introduction of new bugs.

External dependencies that might require python 3 compatibility fixes could be installed from forks without waiting for corresponding maintainers to merge them.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.23 ETH (102.92 USD @ $447.49/ETH) attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/NeuroVault/NeuroVault/630/717).
 * Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $28,438.28 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks, 1 day from now. Please review their questions below:


1. [AllenChong](https://gitcoin.co/profile/AllenChong) has started work. 	
 * Q: PowerKill Project

chrisfilo:
@AllenChong "PowerKill Project"? What do you mean?

abitrolly:
A can of worms. :)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 4 weeks from now. Please review their questions below:


1. [eayoungs](https://gitcoin.co/profile/eayoungs) has started work. 	
[...]

PR #635, opened 2018-07-02
Dockerfile cleanup

abitrolly:
#630 research
- npm is not needed for nginx (5f661537)
- apt-get update was already run
- [WORKDIR](https://docs.docker.com/v17.09/engine/reference/builder/#workdir) creates its dirs automatically

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,7 +73,7 @@ jobs:
         command: if [ "$CIRCLE_NODE_INDEX" == "0" ]; then curl --retry 10 --retry-delay 15 -vkf http://localhost; fi
     - run:
         name: set -o pipefail && docker-compose run uwsgi python manage.py test -v 2 2>&1 | tee log.txt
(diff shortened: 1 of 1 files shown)


chrisfilo:
We are still dependent on coffeescript - this is probably why your tests fail.

abitrolly:
@chrisfilo looked for coffeescript, but found only Django tags. Can you point me to coffeescript code?

chrisfilo:
It might be that those tags is all that's left over, but that's enough to trigger the error. You would have to remove the tags and the mention of coffeescript in settings.

chrisfilo:
BTW - thanks for helping to clean this up!

abitrolly:
In fact I came for the bounty in #630, but I could not promise to do anything until I know that it is real. May need to much time to learn the codebase, and while I learn it, cleaning the things up is a natural thing to make this time worthy.

chrisfilo:
I see - thanks for clarifying. I can assure you that the bounty is very real.

abitrolly:
This now depends on #641, but I have to admit that my abilities for CoffeeScript are rather limited.

PR #640, opened 2018-07-03
Specify where the AHBA data is located

abitrolly:
As I understand there is nothing more important in this image. However, I can not give description about what this data is and how it is different from AHBA. I see that it is postprocessed, but it is not clear why.

### DIFF
--- a/ahba_docker/README.md
+++ b/ahba_docker/README.md
@@ -2,3 +2,12 @@ Sources for Docker image with data from Allen Human Brain Atlas
 
 * http://human.brain-map.org/static/download
 * https://hub.docker.com/r/neurovault/ahba/
+
+The data location is `/ahba_data`.
+```
+/ahba_data
+├── [640K]  Richiardi_Data_File_S2.xlsx
+├── [223K]  corrected_mni_coordinates.csv
+├── [1.6M]  probe_info_max1.csv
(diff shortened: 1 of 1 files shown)


chrisfilo:
It comes in multiple compressed CSV files and is converted into single HDF5 file optimized for indexing. More details at https://github.com/NeuroVault/NeuroVault/blob/master/ahba_docker/preparing_AHBA_data.py

abitrolly:
The HDF5 itself is optimized for indexing, or there are specific indexes being built? What are other files along with `.h5`? Why `max1` extension?

chrisfilo:
That's a lot of questions. Maybe it would help if you provided some context about what you are trying to achieve.

abitrolly:
Just trying to figure out the auditory of this image. Allows more people discover it if needed.

abitrolly:
The secondary goal is to decouple main Dockerfile from this image. I am almost sure that Python 3 porting will be impossible on Jessie.

abitrolly:
The auditory of the image are people who need to work with AHBA data. Those people may appreciate when they can download the data already prepared in HDF5.

abitrolly:
No need to make it Python dependent - it could be just an image `FROM scratch` with data that is mounted to main application as volume through `docker-compose`.

PR #646, opened 2018-07-05
Upgrade pip before installing dependencies

abitrolly:
This removes repeated stderr message during container build:

    You are using pip version 8.1.2, however version 10.0.1 is available.
    You should consider upgrading via the 'pip install --upgrade pip' command.


chrisfilo:
Neat!

PR #650, opened 2018-07-06
 Decouple NeuroVault image from AHBA data

abitrolly:
`docker-compose` doesn't run on my SELinux-enabled system, so I rely on CircleCI to test this.


abitrolly:
Test are failing, but I am afraid I am not the one who knows how to fix them. )

PR #656, opened 2018-09-27
fix how cogat api results are dealt with.

rwblair:
Calls to get a single task no longer return an array with a single object, instead just return the object itself.


chrisfilo:
thx

PR #679, opened 2019-04-22
Fix ahba docker image build broken due to jessie updates

chrisgorgo:
The updated image has been pushed to docker hub.


rwblair:
Was able to build dockerfile with out issue and do a `docker-compose up` without issue. 

@chrisgorgo is rebuilding the containers on production urgent?

chrisgorgo:
It's unlikely there is a need to rebuild containers in production.

PR #680, opened 2019-04-24
Fixing schema.org

chrisgorgo:
It turns out identifier needs to be a list.


chrisgorgo:
Replaces by https://github.com/NeuroVault/NeuroVault/pull/686

PR #684, opened 2019-05-06
Pin more dependencies to fix tests

chrisgorgo:


chrisgorgo:
BTW there is no need to rebuild/pull images on the server. This is only so the CI will pass.

chrisgorgo:
I also rebased all of my recent PRs on top of this one so you could easily see that they do not break any tests.

PR #724, opened 2019-08-30
add variance as a map type choice in BaseStatistcsMap model 

### LINKED ISSUE #722: Add "Variance" as a Map Type
We want to upload fitlins / neuroscout single subject results, and for maximal utility in future mega-analyses, variance as a map type would be very useful.

@rwblair seems easy right?

rwblair:
fixes #722

@chrisgorgo any objections to this?

Rebuilding my local environment to test this to see if there are any problems and generate migration.


chrisgorgo:
I don't forsee any issues. You will have to apply db migration in production though.

adelavega:
Lightning quick! Thx!

PR #742, opened 2021-04-21
Fix build error caused by missing `django-mailgun`

abitrolly:
Could not find a version that satisfies the requirement `django-mailgun`


rwblair:
mailgun has been removed as a requirement

PR #767, opened 2022-10-31
[CI] add github actions

### LINKED ISSUE #760: add CI test of docker image building

jdkent:
closes #760


jdkent:
looks like neurovault image is still dependent on pybraincompare to run, is this a package we are keeping?

jdkent:
removed the current ci file and replaced it with the one I was working on.
building and caching work. I had to change the docker build process to make it such that the freesurfer build of neurovault would build after the original neurovault docker image was built (using multistage building). This could change how production would be implemented.

I think there a couple changes to be made before we make this deployable to production
- [ ] tag builds with releases
- [ ] restructure production.yml to use the prebuilt images on this repository

Otherwise this is working/ready for review.

docker-compose.yml
```diff
@@ -4,7 +4,8 @@ services:
     restart: always
     build:
       context: .
-      dockerfile: ./compose/django/Dockerfile_fs
+      dockerfile: ./compose/django/Dockerfile_multi
```
adelavega, line 7: Why do we need 2 Docker files?

adelavega (approved):
LGTM! I'll let ross take a look, but let's merge this soon so we can make other changes

adelavega:
Currently the multistage build i used for CI, but we have not currently adopting it for production

Issue #777, opened 2023-04-11
Synchronize _fs and _multi builds between CI and production


adelavega:

rwblair:
fixed in #783

PR #783, opened 2023-04-26
Use DRF built in token auth, add views to create and regen this token

rwblair:
Still need to purge old oauth2_provider code.


neurovault/apps/users/views.py
```diff
@@ -80,155 +82,15 @@ def password_change_done(request):
...
-        token_list = self._refresh_token_queryset(self.request.user, self.object.id)
-
-        for refresh_token in token_list:
-            refresh_token.revoke()
-
-        messages.success(self.request, self.success_message)
-
-        return HttpResponseRedirect(success_url)
+@login_required
+def view_token(request, regenerate=False):
```
adelavega, line 86: Should we allow people to view tokens that were previously generated? Or be like Github that you can only vie when generating?

adelavega:
Does the fact that its built in mean that this already will authorize the API?

neurovault/apps/users/views.py
rwblair: I'll leave them shown for now, can get more sophisticated in the future.

rwblair:
> Does the fact that its built in mean that this already will authorize the API?

Added test to verify this.

[9379139](https://github.com/NeuroVault/NeuroVault/pull/783/commits/9379139df5e546e8e6126245463d457a4fc7a6aa) Fixes #777

PR #806, opened 2023-08-10
re-add gene expression routes and views.

rwblair:
I believe these were removed originally because we couldn't get the ahba docker image to build. I was able to get the preparing_AHBA_data.py script to run, but the df.drop removing coordinates outside of the brain caused key errors. changing the drop to use errors='ignore' allowed script to run to completion and produced a store_max1_reduced.h5 that was about the same size as the old one, but according to h5diff contained many differences. Opting to use old version of it.


rwblair:
upgrading to nibabel 5.1 causes issues with nidmresults and nidmresults-fsl. Looking into patching these.

rwblair:
Dropping down to nibabel 4 gets all but one of the tests to pass. Issues are around nidmresults which need to be fixed on their own anyways.

Issue #600, opened 2017-10-20
Mismatch in number of public collections


tyarkoni:
The number of public collections listed at [https://neurovault.org/collections/stats](https://neurovault.org/collections/stats) doesn't match what's visible in the table under "see all collections"--possibly because the former counts empty collections and the latter doesn't (per @chrisfilo's speculation). Probably makes sense to drop empty collections from all counts on the stat page.

PR #819, opened 2025-01-24
Advanced collection search.

rwblair:
- Adds filters for DOI, map type, modality, and task to collection table. 
- Adds new last modified column.
- Does not show empty collections in public view.
- Merges my collections and public collections templates.
- Removes "add new collection" button from bottom of page.
- Fixes delete Image 500 error from bad redirect.

We need look at the performance of the table JSON endpoint in production. This adds some potentially costly database operations.


rwblair:
fixes #600 
fixes #788

PR #822, opened 2025-01-24
Streamline Image Upload + Misc

adelavega:
- Fix "upload folder" (backend and frontened issues)
- Refactor image forms to use crispy layout
- Redirect straight to image editor from upload dialog
- Add nav to image editor (next / back)
- Streamlining of image upload form
- Add functionality to auto fill in meta-data from previously filled out images


adelavega:
Okay, current the issue I'm facing is that to display the file in a simple way that is read only, the validation logic fails (because it doesnt get a file back)

e.g.
```
       self.fields["file"].widget = forms.TextInput(
            attrs={"readonly": True, "class": "form-control"}
 ```
 
 In general the validation logic is difficult to work with and doesn't make much sense in the context of editing in bulk so two options:
 
 1) Keep it simple and leave file selector as it was before (changable)
 2) OR Move all the file validation logic into a separate place, and only trigger it in the scenario when you are adding a new file from scratch (e.g. one by one).
 3) We could potentially disable adding files one by one and effectively funnel everyone through upload by folder (with validation happening earlier), and then editing laer.

PR #825, opened 2025-01-30
copy metadata correctly

jdkent:


jdkent:
todo
- N/A selection not always respected (represented)
- the radio buttons are not always updated visually (sometimes happens, sometimes does not)

adelavega:
- Set default for new images to ------
- By default only show images with valid meta-data 
- Maybe not all fields should be copied?

PR #833, opened 2025-02-03
ENH: Streamline collection creation

adelavega:


neurovault/apps/statmaps/migrations/0002_statisticmap_cognitive_paradigm_name_and_more.py
jdkent, line 1: I'm not as familiar with django, but is this analogous to the migrations kept track with alembic? If so, do we want to keep this file for the history of how the database has changed?

adelavega: Correct. Yep we want to but I messsed something up so I delete all the migration I've made and remade them into a single migration which is now tracked.

jdkent:
- self.helper.form_action in forms.py (does this do anything?) 
- redirect is not bringing me to the wizard
- progress bar for second upload is not showing, and does not appear to upload images.

adelavega:
- Image form is also acting up especially when no task is selected

adelavega:
@jdkent ready for another review

neurovault/apps/statmaps/forms.py
```diff
@@ -79,12 +79,23 @@
...
             "fields": [
                 "name",
+                Field(
+                    "publication_status",
+                    template="statmaps/fields/toggle_radio_field.html",
+                ),
                 "DOI",
-                "communities",
+                "preprint_DOI",
+                HTML("<div class='alert alert-warning' role='alert' id='not_intended_alert'>Collections not associated with a publication or preprint are not guaranteed long-term archival.</div>"),
```
jdkent, line 88:
```suggestion
                HTML("<div class='alert alert-warning' role='alert' id='not_intended_alert'>Collections not associated with a publication are not guaranteed to be retained for long term archival .</div>"),
```

PR #855, opened 2025-10-27
Fix neurosynth upload

adelavega:
attempt to make simple upload form validate

@rwblair I can't get neurovault docker to work locally, give this a shot?


rwblair:
Just ended up needing the cognitive atlas layout from the main stat map form.

